PR #190, opened 2021-05-26
added test report coverage

galmail:
This PR includes test reports with coverage, which is generated as an html inside the **test-report** folder


jest.config.js
```diff
@@ -14,5 +14,21 @@ module.exports = {
     '/tests/e2e/',
     '/build/',
   ],
-  reporters: ['default', 'jest-junit'],
+  coverage: true,
```
jensdev, line 17: I think this should be `collectCoverage`

package.json
```diff
@@ -22,7 +22,7 @@
     "add-suffix": "node ./node_modules/npm-version-suffix/run-add-suffix.js",
     "remove-suffix": "node ./node_modules/npm-version-suffix/run-remove-suffix.js",
     "prepare": "yarn build && husky install",
-    "test": "jest",
+    "test": "jest --coverage",
```
jensdev, line 25: This is not needed

jest.config.js
galmail: oh yes you're right!! I changed it now!

package.json
galmail: ah yes! with the flag set in the jest config, this is not needed. I've removed it now.

jensdev (approved):

PR #205, opened 2021-06-01
Chore: created find helpers

galmail:


package.json
```diff
@@ -13,7 +13,7 @@
   "scripts": {
     "clean": "rimraf build",
     "start": "yarn test",
-    "start:watch": "yarn test --watch",
+    "start:watch": "yarn test --watch --reporters=\"default\"",
```
jensdev, line 16: Why is this?

package.json
```diff
@@ -69,6 +69,16 @@
     "outputDirectory": "test-results",
     "outputName": "jest-junit.xml"
   },
+  "jest-stare": {
```
jensdev, line 72: Why is this?

src/services/data/comments.ts
```diff
@@ -62,6 +62,51 @@ export default (client, httpAuth: HttpInstance) => ({
...
+   * @param schemaId the schema Id
+   * @param documentId the document Id
+   * @param rql an optional rql string
+   * @returns the first element found
+   */
+  async findById(
+    id: ObjectId,
+    schemaId: ObjectId,
+    documentId: ObjectId,
+    rql?: RQLString
```
jensdev, line 77: Optional parameters should go in seperate object.

src/services/data/comments.ts
```diff
@@ -62,6 +62,51 @@ export default (client, httpAuth: HttpInstance) => ({
...
+  ): Promise<Comment> {
+    const rqlWithId = rqlBuilder(rql).eq('id', id).build();
+    const res = (
+      await client.get(
```
jensdev, line 81: Any preference to do it like vs for example something like:
```ts
  async findById(id: ObjectId, rql?: RQLString): Promise<Schema> {
    const rqlWithId = rqlBuilder(rql).eq('id', id).build();
    const res = await this.find({ rql: rqlWithId});
    return res.data[0];
  },
```

src/services/dispatchers/index.ts
```diff
@@ -15,8 +15,10 @@ export const dispatchersService = (
     basePath: DISPATCHERS_BASE,
   });
 
+  const dispatchersMethods = dispatchers(client, httpWithAuth);
+
   return {
-    ...dispatchers(client, httpWithAuth),
+    ...dispatchersMethods,
```
jensdev, line 21: Any reason why this is changed?

package.json
galmail: I had to change it, cause the tests were going into infinite loop!!

src/services/data/comments.ts
galmail: I wasn't sure if we should... as the find helpers should be kept simple.

src/services/data/comments.ts
galmail: that's what I wanted to do, but unfortunately `this` keyword refers to the calle.

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [3.1.0]
+
+- Added `findById`, `findByName` and `findFirst` methods to services
+
+## [3.0.1]
```
jensdev, line 14: Remove this one

src/services/data/comments.ts
galmail: all done! thanks!

src/services/data/comments.ts
galmail: done!

CHANGELOG.md
galmail: ok done

jensdev (approved):

PR #231, opened 2021-06-10
Feature: Profiles Service

galmail:
This PR includes all the endpoints for profiles service.
NOTE: there seems to be an inconsistency in Swagger, in the DELETE methods for Logs Service and Medications Service. Both should return AffectedRecords to be consistent with the rest of the API.


src/services/profiles/profiles.ts
```diff
@@ -0,0 +1,100 @@
...
+   * Permission | Scope | Effect
+   * - | - | -
+   * none | | View your profile
+   * none | `staff enlistment` | View all the profiles of the group
+   * `VIEW_PATIENTS` | `global` | View all profiles
+   *
+   * @param rql an optional rql string
+   * @returns PagedResult<Profile>
+   */
+  async find(options?: { rql?: RQLString }): Promise<PagedResult<Profile>> {
```
jensdev, line 18: Add `findById` and `findFirst`

src/client.ts
```diff
@@ -13,6 +13,7 @@ import {
   mailsService,
   dispatchersService,
   paymentsService,
+  profilesService,
```
jensdev, line 16: Add link to README as well

src/services/profiles/types.ts
```diff
@@ -0,0 +1,107 @@
...
+  impediments?: Array<Impediments>;
+  medication?: Array<Medication>;
+  groups?: Array<Group>;
+  customFields?: Record<string, string>;
+  fibricheckInfo?: string;
+  creationTimestamp?: Date;
+  updateTimestamp?: Date;
+}
+
+export type ProfileCreation = Pick<
```
jensdev, line 32: When I call sdk.profiles.create with an empty object. I get `required_fields: [ 'birthday', 'country', 'gender', 'id' ]`

src/services/profiles/profiles.ts
galmail: ah good spot! I also added these now to the payments services.

src/client.ts
galmail: ah yes, I forgot! I added it now!

src/services/profiles/types.ts
galmail: erm so what exactly are you expecting? I'm not sure what are the required fields..

jensdev: Check the swagger docs. https://developers.extrahorizon.io/swagger-ui/?url=https://developers.extrahorizon.io/services/profiles-service/1.1.3/openapi.yaml#/Profile/post_ 
Only `region` is not required according to the swagger

src/services/profiles/groups.ts
```diff
@@ -0,0 +1,100 @@
...
+   * `ADD_PATIENT` | `staff enlistment` | Add a group enlistment for any profile of this group
+   * `ADD_PATIENT` & `ACTIVATE_PRESCRIPTIONS` | `global` | Add a group enlistment for any profile for any group
+   *
+   * @param profileId Id of the targeted profile
+   * @param requestBody Group data
+   * @returns Group
+   * @throws {ResourceAlreadyExistsError}
+   * @throws {ResourceUnknownError}
+   */
+  async addGroupToProfile(
```
jensdev, line 20: `sdk.profiles.groups.addToProfile` instead of `sdk.profiles.groups.addGroupToProfile` ?

src/services/profiles/groups.ts
```diff
@@ -0,0 +1,100 @@
...
+    profileId: ObjectId,
+    requestBody: Group
```
jensdev, line 22: Needs specific type. `group_id` is a required field

src/services/profiles/logs.ts
```diff
@@ -0,0 +1,107 @@
...
+   * @param profileId Id of the targeted profile
+   * @param groupId Id of the targeted group
+   * @param requestBody CommentBean
+   * @returns LogEntry
+   * @throws {ResourceUnknownError}
+   */
+  async create(
+    profileId: ObjectId,
+    groupId: ObjectId,
+    requestBody: CommentBean
```
jensdev, line 22: No beans

src/services/profiles/types.ts
galmail: yes you're right, I missed that. So I've changed it, have a look and let me know. Thanks!

src/services/profiles/groups.ts
galmail: yeah agree! sounds better!

src/services/profiles/logs.ts
galmail: ha ok!

src/services/profiles/groups.ts
galmail: ah yes, groupId is mandatory! I've now changed it.

src/services/profiles/logs.ts
```diff
@@ -0,0 +1,113 @@
...
+   * - | - | -
+   * `VIEW_PROFILE_LOG_ENTRIES` | `staff enlistment` | Retrieve a list of log entries for any profile of this group
+   * `VIEW_PROFILE_LOG_ENTRIES` | `global` | Retrieve a list of log entries for any profile of any group
+   *
+   * @param profileId Id of the targeted profile
+   * @param groupId Id of the targeted group
+   * @returns PagedResult<LogEntry>
+   * @throws {ResourceUnknownError}
+   */
+  async getLogs(
```
jensdev, line 51: Hmm now it's `sdk.profiles.logs.getLogs` 
But `find` also sounds misleading as you can not pass in any filter.
Maybe `getAll` ? or `list` ?

jensdev: On a closer look, it does seem as you can pass in RQL. The result is also paged, so you would need at least be able to pass in `limit`. So just call it `find` ?

src/services/profiles/medications.ts
```diff
@@ -0,0 +1,68 @@
...
+export default (client, httpAuth: HttpInstance) => ({
+  /**
+   * Retrieve a list of all the defined medications
+   * Permission | Scope | Effect
+   * - | - | -
+   * none | | Everyone can use this endpoint
+   *
+   * @returns PagedResult<Medication>
+   */
+  async getMedications(): Promise<PagedResult<Medication>> {
```
jensdev, line 20: Maybe `list` ?
It's an atypical PageResult as it just returns everything (138 strings) with no limit

src/services/profiles/medications.ts
```diff
@@ -0,0 +1,68 @@
...
+   * `UPDATE_PROFILES` | `global` | Add a new medicine to a profile for any group
+   *
+   * @param profileId Id of the targeted profile
+   * @param requestBody Medication data
+   * @returns Medication
+   * @throws {ResourceUnknownError}
+   */
+  async create(
+    profileId: ObjectId,
+    requestBody: Medication
```
jensdev, line 38: `medication_frequency` is a required field, but it's not correctly shown in swagger

jensdev: I think number and unit are also required, but not in swagger and also not in the response. But if you do not pass them in -> 500 error

src/services/profiles/medications.ts
galmail: I was thinking about `list` but that means we need to change it everywhere (in the other services) so I left it as is for now, but we can ask in the call today, and then create an issue to update all services

src/services/profiles/logs.ts
galmail: > Hmm now it's `sdk.profiles.logs.getLogs`
> But `find` also sounds misleading as you can not pass in any filter.
> Maybe `getAll` ? or `list` ?

agree, see my comment below

galmail: > On a closer look, it does seem as you can pass in RQL. The result is also paged, so you would need at least be able to pass in `limit`. So just call it `find` ?

Where do you see the RQL?

jensdev (approved):

PR #236, opened 2021-06-14
Feature: Notifications service

galmail:
This PR includes all the endpoints for the notifications service
NOTE: I noticed that the error codes are missing

### DIFF
--- a/src/services/notifications/index.ts
+++ b/src/services/notifications/index.ts
@@ -1,0 +1,18 @@
+import type { HttpInstance } from '../../types';
+import httpClient from '../http-client';
+import notifications from './notifications';
+import { NOTIFICATIONS_BASE } from '../../constants';
+
+export type NotificationsService = ReturnType<typeof notifications>;
+
+export const notificationsService = (
+  httpWithAuth: HttpInstance
+): NotificationsService => {
+  const client = httpClient({
+    basePath: NOTIFICATIONS_BASE,
+  });
+
+  return {
+    ...notifications(client, httpWithAuth),
+  };
+};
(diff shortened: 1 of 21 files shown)


src/services/notifications/notifications.ts
```diff
@@ -0,0 +1,146 @@
...
+      .data;
+  },
+
+  /**
+   * Find By Id
+   * @param id the Id to search for
+   * @param rql an optional rql string
+   * @returns the first element found
+   */
+  async findAllById(
```
jensdev, line 88: Might sound confusing? Maybe we can discuss naming during call? In other services there's just one 1 endpoint for your own / others

src/services/notifications/index.ts
```diff
@@ -0,0 +1,25 @@
...
+export type NotificationsService = ReturnType<typeof notifications> &
+  ReturnType<typeof health> & {
+    settings: ReturnType<typeof settings>;
+  };
+
+export const notificationsService = (
+  httpWithAuth: HttpInstance
+): NotificationsService => {
+  const client = httpClient({
+    basePath: NOTIFICATIONS_BASE,
```
jensdev, line 17: Add `    transformRequestData: decamelizeKeys`

src/services/notifications/notifications.ts
galmail: cool, has been removed now following the discussion with Kobe

src/services/notifications/index.ts
galmail: ah right, done!

jensdev (approved):

PR #325, opened 2021-07-30
Chore: OAuth1 authenticate returns tokenData

galmail:


src/http/oauth1.ts
```diff
@@ -125,6 +125,7 @@ export function createOAuth1HttpClient(
         secret: tokenResult.data.tokenSecret,
       });
     }
+    return tokenData;
```
ThomasReyskens, line 128: it should return the token result, this is just the data that you provide as input, no?

galmail: if the user provides it, then yes, otherwise it will generate the tokenData and it will return it.

jensdev: There's 2 ways to auth with oauth1 atm. Token and password. If you use token, it's always the same token that's put on to the calls. If you use password, you get new tokens from the /tokens endpoint. Unless you have 2FA enabled. Then you need to use `confirmMfa` to get tokens. So we should return those there as well?

ThomasReyskens (dismissed):

jensdev (approved):

PR #378, opened 2021-08-31
Feature/find all generator

jensdev:
image


src/version.ts
```diff
@@ -1 +1 @@
-export const version = '4.3.1';
+export const version = '4.5.0';
```
ThomasReyskens, line 1: did we skip 4.4?

jensdev: Hmm, no I think I forgot to run `yarn build` last time. The file should be ok in the published packages because there we run `yarn build` prepublish

ThomasReyskens (approved):

src/version.ts
ThomasReyskens: ok!

ThomasReyskens (approved):

Issue #424, opened 2021-09-24
userId getter not working on the oauth1/oauth2 client in RN


jensdev:
## Description

The userId getter is not working in RN. Becomes apperent when trying to create a message notification. 
RN does have support for getters and it does work when adding it with

```typescript
Object.defineProperty(httpWithAuth, 'userId', {
    get() {
       return tokenData?.userId;
  });
```

We could also just use regular function in stead of get.

## Code example

```typescript
const notification = await sdk.notifications.create({
        userId: userId,
        important: false,
        type: 'message',
        fields: {
          title: 'message title',
          body: 'this the a body text',
        },
      });
```

## Specifications

- Version: v5.0.0
- Platform: React-Native
- Endpoint or service affected: notifications

PR #462, opened 2021-11-10
Refactor getter for RN

jensdev:
#424 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get#defining_a_getter_on_existing_objects_using_defineproperty


ThomasReyskens:
Add some comments why it is done like that, else other people are going to hate and refactor 😄

jensdev:
True, I've added a note :)

ThomasReyskens (approved):

PR #706, opened 2022-12-06
Turn off maxBodyLength and redirects

AaronTrippaers:


wwaaijer-exh (commented):
Looks good, but can it be that the issue in not mentioned in the commit message? (I would go for the GH issue)

wwaaijer-exh (approved):

PR #732, opened 2023-03-10
Removed the need for another intermediate auth data model

wwaaijer-exh:


src/client.ts
```diff
@@ -254,8 +155,7 @@ export function createClient<T extends ClientParams>(rawConfig: T): Client<T> {
     auth: ('authenticate' in httpWithAuth && httpWithAuth.authenticate
       ? {
           ...authService(httpWithAuth),
-          authenticate: (oauth: AuthParams) =>
-            httpWithAuth.authenticate(parseAuthParams(oauth)),
```
wwaaijer-exh, line 258: This was basically the only thing I wanted to change, splitting up the `parseAuthParams` 😬

src/http/oauth1.ts
```diff
@@ -2,13 +2,12 @@ import axios from 'axios';
 import OAuth from 'oauth-1.0a';
 import { HmacSHA1 } from 'crypto-es/lib/sha1';
 import { Base64 } from 'crypto-es/lib/enc-base64';
-import { ParamsOauth1 } from '../types';
+import { Oauth1AuthParams, ParamsOauth1 } from '../types';
```
AaronTrippaers, line 5: Would `oAuth1AuthenticateParams` and `Oauth1ClientParams` make more sense here? Seeing those next to each other is really weird.

If we even can change the names without breaking backwards compatibility

AaronTrippaers: The same is true for oAuth2 file

AaronTrippaers (approved):

PR #733, opened 2023-03-10
Feature/svcs 337 sdk supports OIDC

chad-pelly:


chad-pelly:
I assume the user used for testing is missing the permission.

src/models/oidc/Providers.ts
```diff
@@ -0,0 +1,31 @@
+export interface BaseOidcProvider {
```
chad-pelly, line 1: These are extractions from other services. I assume we have a reference repository ticket somewhere to some day set up a joint models package?

src/services/auth/oidc.ts
```diff
@@ -0,0 +1,60 @@
+import { HttpInstance, OidcService } from '../../types';
+import { HttpClient } from '../http-client';
+
+export default (
+  oidcClient: HttpClient,
+  httpWithAuth: HttpInstance
+): OidcService => ({
+  async createProvider(requestBody) {
+    const { data } = await oidcClient.post(
```
chad-pelly, line 9: The other endpoints don't extract `{ data }` they just return x.data. This is more of a preference for me and I can quickly change it back if we don't like it. Simple code yamsayin'.

src/models/oidc/Providers.ts
chad-pelly, line 1: I also feel that the `types` file referencing all service interfaces and all models is getting cumbersome. Here I have started the implementation as if were a traditional service.

Alternative I can put it all in the `types` file and create a reference repository ticket to change that at a later date.

tests/unit/auth/oidc.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  beforeAll(async () => {
+    await sdk.auth.authenticate({
+      email: process.env.API_USERNAME,
+      password: process.env.API_PASSWORD,
+    });
+  });
+
+  beforeEach(async () => {
+    const data = createProviderData();
```
chad-pelly, line 22: This technically would be nicer if it was declared individually in each test, especially the create. But it seemed unnecessary.

src/services/auth/types.ts
```diff
@@ -434,3 +439,61 @@ export interface AuthOauth1Service {
...
+
+  /**
+   * Update an OpenId Connect Provider
+   * Permission | Scope | Effect
+   * - | - | -
+   * `UPDATE_OIDC_PROVIDER` | `global` | **Required** for this endpoint
+   */
+  updateProvider(
+    providerId: string,
+    requestBody: Partial<OidcProviderUpdate>
```
AaronTrippaers, line 473: (minor) The OidcProviderUpdate is already a partial, so this is technically not need 😄

tests/unit/auth/oidc.ts
```diff
@@ -0,0 +1,91 @@
...
+    provider = await sdk.auth.oidc.createProvider(data);
+  });
+
+  afterAll(async () => {
+    await sdk.auth.oidc.deleteProvider(provider.id);
+  });
+
+  it('Creates an OpenID Connect provider', async () => {
```
AaronTrippaers, line 30: I would still do this create a call in the test here, now the test and the beforeEach are mangled.

src/models/oidc/Providers.ts
wwaaijer-exh: Yes! The idea was to generate the types based on the swagger files some day. Same goes for the swagger files themselves, generating them based on the types/validators

src/models/oidc/Providers.ts
```diff
@@ -0,0 +1,31 @@
...
+  name: string; // Human friendly name of the provider, which can also be used in the oidc login urls
```
wwaaijer-exh, line 2: (minor) If you move these comments to JSDoc they'll nicely show up in the editor for user of the SDK. Like:
```
/** Human friendly name of the provider, which can also be used in the oidc login urls */
name: string;
```

src/services/auth/oidc.ts
wwaaijer-exh: Yeah, this is way better 👍 I've a note for this as well. Idea right now was to refactor this away to be included in the `HttpClient` logic, as this is happening in basically every file 😬

tests/__helpers__/oidc.ts
```diff
@@ -0,0 +1,23 @@
...
+import { OidcProviderCreation } from '../../src/models/oidc/Providers';
+
+export const generateRandomName = () =>
+  `${crypto.randomBytes(5).toString('hex')}-${crypto
+    .randomBytes(5)
+    .toString('hex')}`;
+export const createProviderData = (
+  data?: Partial<OidcProviderCreation>
+): OidcProviderCreation => ({
+  name: `${crypto.randomBytes(5).toString('hex')}-${crypto
```
wwaaijer-exh, line 11: I assume you can reuse `generateRandomName` here? 😄

tests/unit/auth/oidc.ts
```diff
@@ -0,0 +1,91 @@
+import { createClient, rqlBuilder } from '../../../src';
+import { createProviderData, generateRandomName } from '../../__helpers__/oidc';
+import { OidcProviderResponse } from '../../../src/models/oidc/Providers';
+
+describe('Enables an OpenID Connect workflow', () => {
```
wwaaijer-exh, line 5: These tests seem to be written as end 2 end tests, not unit tests

wwaaijer-exh (approved):
Nice nice 👍

PR #736, opened 2023-03-23
Feature/svcs 790 retroactively update inline documentation

KobeLeysen:
Love to get your feedback! A first inline documentation tryout

*Code style an quality assurance jobs are failing because of unused imports. But this is incorrect as the imports are being used in JSDoc and the linter doesn't use this.


KobeLeysen:
Looking into: https://github.com/eslint/eslint/issues/9050 as we don't want to import just for the sake of JSDoc

wwaaijer-exh (approved):

PR #741, opened 2023-03-28
SVCS-802 Accept oAuth1/2 tokens in config

wwaaijer-exh:
Might be easier to review the commits separately.
While making these changes it became clear some tests were either incomplete or misleading and needed some refactoring.

### DIFF
--- a/src/http/oauth2.ts
+++ b/src/http/oauth2.ts
@@ -24,7 +24,14 @@ export function createOAuth2HttpClient(
   http: HttpInstance,
   options: ParamsOauth2
 ): OAuth2HttpClient {
-  let tokenData: TokenDataOauth2;
+  let tokenData: Partial<TokenDataOauth2>;
+
+  if ('refreshToken' in options && 'accessToken' in options) {
+    tokenData = {
+      refreshToken: options.refreshToken,
+      accessToken: options.accessToken,
+    };
+  }
 
   const clientCredentials = getClientCredentials(options);
 
@@ -103,7 +110,7 @@ export function createOAuth2HttpClient(
     ) {
       originalRequest.isRetryWithRefreshedTokens = true;
       await refreshTokens();
-      return http(originalRequest);
+      return httpWithAuth(originalRequest);
     }
 
     throw error;
--- a/tests/unit/client/oauth2.test.ts
+++ b/tests/unit/client/oauth2.test.ts
@@ -18,9 +18,9 @@ const mockParams = {
   freshTokensCallback: freshTokensCallbackMock,
 };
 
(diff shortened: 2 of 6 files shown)


src/http/oauth2.ts
```diff
@@ -103,7 +110,7 @@ export function createOAuth2HttpClient(
     ) {
       originalRequest.isRetryWithRefreshedTokens = true;
       await refreshTokens();
-      return http(originalRequest);
+      return httpWithAuth(originalRequest);
```
wwaaijer-exh, line 113: This actually solved a bug introduced in the current version we're working on. Was caught after correcting the test suite, so was definitely worth it already 😬

tests/unit/client/oauth2.test.ts
```diff
@@ -244,14 +288,23 @@ describe('OAuth2HttpClient', () => {
...
-      await confidentialHttpWithAuth.extraAuthMethods.authenticate(mockAuth);
+      await confidentialHttpWithAuth.extraAuthMethods.authenticate(
+        emailAuthData
+      );
+    expect(authenticateResult).toStrictEqual({ accessToken: mockToken });
+
+    // Expect successive requests to be authenticated
     nock(mockParams.host).get('/test').reply(200, '');
 
     const result = await confidentialHttpWithAuth.get('test');
```
chad-pelly, line 305: (minor) Could we randomHexString / uuid or something: `mockToken`, we have `test`, `/test` and `test` and it makes the context a little harder to read 🤔 

Same for clientId and secret, could we make them the same references if they're related objects?

tests/unit/client/oauth2.test.ts
```diff
@@ -204,34 +245,37 @@ describe('OAuth2HttpClient', () => {
         },
       });
 
+    let mfaError;
+    try {
```
chad-pelly, line 249: (opinion / ignore) Just for cleanliness  as this is a large test, I would 
`await httpWithAuth.extraAuthMethods.authenticate(emailAuthData).catch(error => {
      mfaError = error;
 });`

chad-pelly (approved):
Looks great, a few opinions / minors, but nothing blocking at all 👌

tests/unit/client/oauth2.test.ts
wwaaijer-exh: 💯 Will update

chad-pelly (approved):
:white_check_mark: :saluting_face: :muscle::skin-tone-3: :+1::skin-tone-3: :heavy_check_mark: :ballot_box_with_check: :rocket:

PR #742, opened 2023-03-28
SVCS-490 Added errors and inline documentation supporting OIDC

wwaaijer-exh:


src/errors.ts
```diff
@@ -183,6 +183,11 @@ export class DefaultLocalizationMissingError extends BadRequestError {}
 export class IDFormatError extends BadRequestError {}
 export class ProfileAlreadyExistsError extends BadRequestError {}
 
+export class OAuth2MissingClientCredentialsError extends BadRequestError {}
+export class InvalidNonceError extends BadRequestError {}
+export class OidcProviderResponseError extends BadRequestError {}
```
chad-pelly, line 188: (opinion) this feels a little odd to be a bad request error

chad-pelly (approved):
👌

src/errors.ts
wwaaijer-exh: Yes, I can definitely see that. The reasoning there is that this error most likely is due to the misconfiguration of the provider object we have on record filled in by our client. Therefor we choose to make this a 4xx error. (we would like to keep 5xx errors reserved for errors that we need to act upon)

PR #743, opened 2023-03-30
SVCS-801 Added login attempts endpoints 

chad-pelly:


src/services/auth/oidc/loginAttempts/types.ts
```diff
@@ -0,0 +1,72 @@
+import { OptionsWithRql, PagedResultWithPager } from '../../../types';
+import { FindAllIterator } from '../../../helpers';
+
+export interface LoginAttemptsService {
```
chad-pelly, line 4: I took some liberties in formatting the documentation in what I thought in a slightly better way here. Please have a lookie at the result in the SDK and see what you think.

src/services/auth/oidc/loginAttempts/types.ts
```diff
@@ -0,0 +1,72 @@
...
+  providerSubjectId: string;
+  /** The Extra Horizon user id */
+  userId: string;
+  /** An Extra Horizon formatted error for failed requests */
+  error: {
+    name: string;
+    message: string;
+    code: number;
+  };
+  creationTimestamp: Date;
```
chad-pelly, line 71: Do we want to add documentation for the normal returned fields `id`, `creationTimestamp` etc... ? I'm not sure that they need explanation let alone multiple copy pastes.

wwaaijer-exh: `id` would indeed be one we could skip. In most cases it could not hurt to document the `creationTimestamp` though. In this case for instance, it might be good to make it extra clear and specify that this is indeed "the time the login attempt happened".

wwaaijer-exh (dismissed):
Looking good 👍

AaronTrippaers (dismissed):

AaronTrippaers (approved):

PR #747, opened 2023-04-07
SVCS-559 Corrected the inline documentation syntax for some titles

wwaaijer-exh:
If the `#` are attached to the text it won't render properly in the VS Code Markdown renderer


chad-pelly (approved):
🔥 🔥 🔥 🔥

PR #749, opened 2023-04-12
Feature/svcs 893 view and configure the user service email templates

chad-pelly:


src/services/users/types.ts
```diff
@@ -348,6 +348,13 @@ export interface UserRoles {
   roles: ObjectId[];
 }
 
+export interface EmailTemplates {
```
chad-pelly, line 351: All instances of EmailTemplates could be changed to EmailTemplateConfigurations if we wanted to be verbose.

tests/unit/users/usersService.test.ts
```diff
@@ -321,4 +321,33 @@ describe('Users Service', () => {
...
+
+  it('Should allow a user with permissions to set the email template configurations', async () => {
+    // The user service will snake case the body
+    const templates = {
+      activationEmailTemplateId: randomHexString(24),
+      reactivationEmailTemplateId: randomHexString(24),
+      oidcUnlinkEmailTemplateId: randomHexString(24),
+    };
+
+    nock(`${host}${USER_BASE}`).put(`/email_templates`).reply(200, templates);
```
wwaaijer-exh, line 348: Is it correct that the request body is missing here? Also, I think we expect the user service the user service to receive and respond with the property names in snake case right?

src/services/users/types.ts
wwaaijer-exh: Fine for me like this 👍

wwaaijer-exh (commented):
👍 1 comment about the tests. Happy to elaborate/discuss

tests/unit/users/usersService.test.ts
chad-pelly: Pogchamp!

tests/unit/users/usersService.test.ts
```diff
@@ -321,4 +321,47 @@ describe('Users Service', () => {
 
     expect(result.minimumLength).toBeDefined();
   });
+
+  it('Should allow a user with permissions to view the email template configurations', async () => {
+    // The user service will respond in snake case and the SDK will convert the response to camel case
+    const data = {
```
chad-pelly, line 327: T'is intentional to only provide three fields instead of four as the request / response is technically partial.

wwaaijer-exh (approved):
Looks good to me 👍 (You're missing the 4th template id in the GET test case now though 😉)

chad-pelly:
> Looks good to me 👍 (You're missing the 4th template id in the GET test case now though 😉)

I've added a partial put request to remove the confusion with why the other tests didn't have all four parameters.

PR #752, opened 2023-04-14
Feature/svcs 894 expose the ex h error as a field in the o auth2 errors

chad-pelly:


wwaaijer-exh (dismissed):
Looking good 👌

wwaaijer-exh (approved):
Even better 👍

chad-pelly (commented):
Came digging for coal and found 🥇

PR #759, opened 2023-06-14
SVCS-1646 Fix permission names for VIEW_GROUP and ADD_ROLE_PERMISSION

AaronTrippaers:


src/services/users/types.ts
wwaaijer-exh, line 1: Just for our convenience during the release of the next version, can you already add an entry to the change logs? 🙏 (might be a while before we take the time to release)

wwaaijer-exh (approved):

PR #761, opened 2023-06-27
Feature/svcs 1702 provide functions to call api function

chad-pelly:


src/services/tasks/api/index.ts
```diff
@@ -0,0 +1,76 @@
+import { ApiService } from './types';
+import { HttpInstance } from '../../../http/types';
+import { HttpClient } from '../../http-client';
+import { OptionsBase } from '../../types';
+
+export default (client: HttpClient, httpAuth: HttpInstance): ApiService => ({
+  async get<T>(name: string, path: string, options: OptionsBase): Promise<T> {
+    const { data } = await client.get(
+      httpAuth,
+      `/api/${name}/${path}`,
```
chad-pelly, line 10: So for all of these methods the path can either start with a / e.g `/users/:id` or without `users/:id`, it will affect line 10. For now I did it without but I'm super happy that it could change if there's a preference.

chad-pelly:
For now I have left out the OPTIONS and HEADERS within the interface.

At first glance it would appear that a OPTIONS request depending on technology used may respond in different ways, either adding the data in the headers OR the body. This means we would have to alter the behaviour of this method to return the whole response and that's not ideal.

The HEAD request is definitely possible and we can just return the response.headers, however it seems strange that it would live alone as the exception, suggested by Aäron.

I have added the OPTIONS and HEAD requests to the HTTP Client so that they may be used by advanced users as a .raw request.

tests/unit/tasks/api/all.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    const { users } = await exh.tasks.api.delete<UsersResponse>(
+      name,
+      `users/${userData[userData.length - 1].id}`,
+      {}
+    );
+    expect(users).toMatchObject(updatedUsers);
+  });
+
+  it('Updates and existing user', async () => {
```
chad-pelly, line 79: I didn't add tests for the `raw.options` and `raw.head` here as they don't belong. I'm not sure they need explicit tests, but I can certainly add them somewhere if needed.

src/services/tasks/api/index.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  async post<T, U>(
+    name: string,
+    path: string,
+    data: U,
+    options: OptionsBase
+  ): Promise<T> {
+    const response = await client.post(
+      httpAuth,
+      `/api/${name}/${path}`,
```
chad-pelly, line 24: I could also always default the paths to `/`

src/services/tasks/api/index.ts
wwaaijer-exh: Discussed, both!

tests/unit/tasks/api/all.ts
wwaaijer-exh: Discussed 👌

tests/unit/tasks/api/all.ts
```diff
@@ -0,0 +1,95 @@
...
+  UserResponse,
+  randomUser,
+  User,
+  userData,
+} from '../../../__helpers__/api';
+import { randomHexString } from '../../../__helpers__/utils';
+
+describe('Tasks - Functions - API', () => {
+  const host = 'https://api.xxx.extrahorizon.com';
+  const name = 'test';
```
wwaaijer-exh, line 16: (minor) Would `functionName` make sense here? Found it a (very) little bit confusing reading the test cases.

wwaaijer-exh (approved):

src/services/tasks/tasks.ts
```diff
@@ -35,4 +36,6 @@ export default (client: HttpClient, httpAuth: HttpInstance): TasksService => ({
   async cancel(taskId, options) {
     return (await client.post(httpAuth, `/${taskId}/cancel`, {}, options)).data;
   },
+
+  api: api(client, httpAuth),
```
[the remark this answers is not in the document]
chad-pelly: Absolutely I've removed it from the index.ts as we still want it to retain tasks.api. 👌 Noice catch. 🐟

tests/unit/tasks/api/all.ts
```diff
@@ -0,0 +1,95 @@
...
+  it('Retrieves data from an API Function', async () => {
+    nock(`${host}${TASKS_BASE}`)
+      .get(`/api/${name}/users`)
+      .reply(200, { users: userData });
+
+    const { users } = await exh.tasks.api.get<UsersResponse>(name, 'users', {});
+    expect(users).toMatchObject(userData);
+  });
+
+  it('Creates a new user', async () => {
```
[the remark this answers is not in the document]
chad-pelly: TRUEEEEEEEEEEEEEEEEE

wwaaijer-exh (approved):
Looking good 👌

AaronTrippaers (approved):

PR #762, opened 2023-06-28
Feature/svcs 741 add the log service v1.0.0

chad-pelly:


.gitignore
```diff
@@ -1,5 +1,5 @@
 # Logs
-logs
+#logs
```
chad-pelly, line 2: Moot and blocking, uncomment if required in the future.

src/client.ts
```diff
@@ -2,20 +2,21 @@ import { ClientParams, ParamsOauth1, ParamsOauth2, ParamsProxy } from './types';
...
-  dispatchersService,
+  notificationsService,
   paymentsService,
-  localizationsService,
   profilesService,
-  notificationsService,
-  eventsService,
+  tasksService,
+  templatesService,
+  usersService,
```
chad-pelly, line 19: Auto import formatting, can revert if needed

src/services/logs/access/types.ts
```diff
@@ -0,0 +1,44 @@
...
+export interface LogsAccessService {
+  /**
+   * ## Retrieve a list of access logs
+   *
+   * **Global Permissions:**
+   * - `VIEW_ACCESS_LOGS` - Allows a user to view access logs
+   * @param options {@link OptionsWithRql} - Add filters to the requested list
+   * @returns An array of access logs {@link AccessLog AccessLog[]}
+   */
+  find(options?: OptionsWithRql): Promise<AccessLog[]>;
```
chad-pelly, line 12: You're probably thinking to yourself... "Hey Chad, how come this find method isn't using a pager unlike the other find methods in the SDK" and don't worry, I'll tell you.

Whilst the Log Service supports the paging model, it is not implemented on the service side, thus we return just the results queried without any magic.

src/services/logs/index.ts
```diff
@@ -0,0 +1,15 @@
...
+import { LOGS_BASE } from '../../constants';
+import access from './access';
+
+export const logsService = (httpWithAuth: HttpInstance): LogsService => {
+  const logsClient = httpClient({
+    basePath: LOGS_BASE,
+  });
+
+  return {
+    access: access(logsClient, httpWithAuth),
```
chad-pelly, line 13: I added +1 level of abstraction here, so we can add more than just access logs in the future if another collection is added to that service.

wwaaijer-exh (approved):
Very nice! 👌

PR #763, opened 2023-07-03
Feature/svcs 728 add task logs to the sdk

chad-pelly:


src/services/tasks/logs/index.ts
```diff
@@ -0,0 +1,33 @@
...
+    const { data } = await client.get(
+      httpWithAuth,
+      `/${taskId}/logs${options?.rql || ''}`,
+      options
+    );
+    return data;
+  }
+
+  return {
+    async find(taskId: ObjectId, options?: OptionsWithRql): Promise<TaskLog[]> {
```
chad-pelly, line 20: Same issue here as the logs service implementation, hard coded pagination, so this is the limitation for now.

src/services/tasks/tasks.ts
```diff
@@ -38,4 +39,5 @@ export default (client: HttpClient, httpAuth: HttpInstance): TasksService => ({
   },
 
   api: api(client, httpAuth),
+  logs: logs(client, httpAuth),
```
chad-pelly, line 42: For context I hate this interface is called `logs` but it resolves to exh.tasks.logs, which to me is much nicer than exh.tasks.taskLogs

tests/unit/tasks/logs/get.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    const logs = await exh.tasks.logs.find(taskId, { rql });
+    expect(logs).toMatchObject(taskLogs);
+  });
+
+  it('Retrieves the first access log matching an RQL query', async () => {
+    const taskId = randomHexString();
+    const rql = rqlBuilder()
+      .ge('timestamp', '2023-06-01T00:00')
+      .limit(10)
```
wwaaijer-exh, line 34: (minor) Since this endpoint (and the tasks logs endpoint) does not support `limit` in RQL these calls would actually error when used targeting a real ExH cluster. I would propose to remove the limit part in these RQL queries

wwaaijer-exh (commented):
1 minor comment, looking good 👍

wwaaijer-exh (approved):
Looks good to me 👌

PR #764, opened 2023-07-03
SVCS-1704 Add interfaces for API Functions in- and output

AaronTrippaers:


src/services/tasks/types.ts
```diff
@@ -39,6 +39,37 @@ export type TaskInput = Pick<
...
+  body: string;
+  isBase64Encoded: boolean;
+}
+
+export interface ApiFunctionResponseObject {
+  statusCode: number;
+  headers?: Record<string, string>;
+  body?: string;
+  isBase64Encoded?: boolean;
+}
```
wwaaijer-exh, line 71: Like we did for the Auth Service during the OpenID Connect epic, can these properties be described with JSDoc? For example: https://github.com/ExtraHorizon/javascript-sdk/blob/dev/src/services/auth/applications/types.ts#L172-L189
We argued back then that the documentation within the SDK might even be more valuable than the public documentation, especially with regard to more detailed information like the fields.

chad-pelly (approved):
Noice documentation! 👍

wwaaijer-exh (approved):
Nice 👌

PR #766, opened 2023-07-04
Feature/svcs 1703 provide functions to list api requests and api request logs

chad-pelly:


src/services/tasks/apiRequests/types.ts
```diff
@@ -0,0 +1,96 @@
...
+  path: string;
+  /** The Extra Horizon user id */
+  userId: ObjectId;
+  /** The Extra Horizon application id */
+  applicationId: ObjectId;
+  /** The status code of the API Function's response */
+  statusCode: number;
+  /** The time the API Function was executed */
+  timestamp: Date;
+  error: ApiRequestError;
```
chad-pelly, line 66: Add a comment here?

src/services/tasks/apiRequests/types.ts
```diff
@@ -0,0 +1,96 @@
...
+  type: ApiRequestErrorType;
+}
+
+export enum ApiRequestErrorType {
+  INVOCATION = 'invocation',
+  RUNTIME = 'runtime',
+  RESPONSE = 'response',
+}
+
+export enum ApiFunctionRequestMethod {
```
chad-pelly, line 88: Move this boi next to the other API Function types please

tests/unit/tasks/apiRequests/get.ts
```diff
@@ -0,0 +1,32 @@
...
+import { apiRequests } from '../../../__helpers__/apiRequests';
+
+describe('Tasks - API Requests', () => {
+  const host = 'https://api.xxx.extrahorizon.com';
+  const exh = createClient({
+    host,
+    clientId: '',
+  });
+
+  it('Lists API Requests matching an RQL query', async () => {
```
chad-pelly, line 14: Previously I had tested the pagination methods incredibly deeply for the loginAttempts, we agreed it would be unnecessary to do it again as they're just copy paste into the interfaces.

Thus I have just tested the construction / targeting of the correct urls here.

PR #768, opened 2023-07-07
SVCS-1822 Added onUploadprogress handlers for file creation

chad-pelly:


src/services/files/types.ts
```diff
@@ -62,37 +65,41 @@ export interface FilesService {
...
-    options?: OptionsBase & { tags: string[] }
+    fileData: FileDataTypes,
+    options?: FileUploadOptions & { tags?: string[] }
   ): Promise<FileDetails>;
   /**
    * Delete a file
    *
    * AccessLevel | Effect
-   * - | -
+   * - | -`
```
AaronTrippaers, line 102: minor: can be removed?

AaronTrippaers (approved):

src/services/files/types.ts
chad-pelly: ooops must of fat fingered....

src/services/files/files.ts
```diff
@@ -28,6 +28,7 @@ export default (client: HttpClient, httpAuth: HttpInstance): FilesService => ({
 
     return (
       await client.post(httpAuth, '/', formData, {
+        onUploadProgress: options?.onUploadProgress,
```
wwaaijer-exh, line 31: Just to be sure, we're not going to wrap the user provided callback in our own function to only forward the properties we want to share (`total` and `loaded`)?

chad-pelly: This is not intended.... I have destructured the event and declared the parameters in the callback. Awesome catch.

wwaaijer-exh (approved):

PR #770, opened 2023-07-10
SVCS-1828 Refactored execute endpoint to functions service

chad-pelly:


tests/unit/tasks/execute.ts
wwaaijer-exh, line 1: (minor) Would it make sense to also move this to `tasks/functions/execute`. To match the file/logical structure of the source code

wwaaijer-exh (approved):

AaronTrippaers (approved):
Looks good!

tests/unit/tasks/functions/post.ts
wwaaijer-exh, line 1: (minor) `tasks/functions/execute` like we do in our services or `tasks/functions/execute/post`? `tasks/functions/post` would be the function creation endpoint/method

wwaaijer-exh: Totally agree! (Even for the services I might have voted for that option 😄)

wwaaijer-exh (approved):

PR #774, opened 2023-07-13
Make some RegisterUserData type fields optional

christopherhex:
Some fields in the `RegisterUserData` type are marked as required even though they are not required by the register endpoint in the API. 

This PR fixes that, and only requires the required parameters for registering a user.


ThomasReyskens (approved):
lgtm

arnedubois (approved):

chad-pelly:
🔥

PR #786, opened 2023-08-03
Feature/svcs 1962 add the settings endpoints to the sdk

chad-pelly:


src/services/files/settings/types.ts
```diff
@@ -0,0 +1,38 @@
+import { AffectedRecords, OptionsBase } from '../../types';
+
+export interface FileServiceSettingsService {
```
chad-pelly, line 3: Uggo name?

src/services/files/types.ts
```diff
@@ -8,6 +8,8 @@ import type {
 } from '../types';
 import { FileUploadOptions } from '../types';
 
+export * from './settings/types';
```
chad-pelly, line 11: WE CANNE FORGET HIM THIS TIME!

src/services/files/settings/index.ts
```diff
@@ -0,0 +1,19 @@
...
+  httpAuth: HttpInstance
+): FileServiceSettingsService => ({
+  async get(options: OptionsBase) {
+    const { data } = await client.get(httpAuth, `/settings`, options);
+    return data;
+  },
+
+  async update(data: FileServiceSettingsUpdate, options: OptionsBase) {
+    const response = await client.put(httpAuth, `/settings`, data, options);
+    return response.data;
```
AaronTrippaers, line 17: Very minor: but the destructuring is in consistent between both functions

src/services/files/settings/types.ts
```diff
@@ -0,0 +1,38 @@
...
+    data: Partial<FileServiceSettingsUpdate>,
+    options?: OptionsBase
+  ): Promise<AffectedRecords>;
+}
+
+export interface FileServiceSettingsUpdate {
+  disableForceDownloadForMimeTypes: string[];
+}
+
+export interface FileServiceSettings {
```
AaronTrippaers, line 35: Should we add js docs comments here to explain the properties?

AaronTrippaers (commented):
Looks good! Only the jsdocs for the FileServiceSettings types are missing I think.

src/services/files/settings/index.ts
chad-pelly: For the put requests we call the input `data` for consistency meaning that we can destructure the output to data.

Any better suggestions? 🤔

src/services/files/settings/types.ts
chad-pelly: YUP!

src/services/files/settings/types.ts
wwaaijer-exh: `FileSettingsService` would be consistent with the other files 🤷

wwaaijer-exh (dismissed):

AaronTrippaers (approved):

wwaaijer-exh (approved):

PR #797, opened 2023-09-06
SVCS-2025 Normalized the existing dispatcher logic

chad-pelly:


src/services/dispatchers/dispatchers/types.ts
```diff
@@ -0,0 +1,83 @@
...
+   */
+  remove(
+    dispatcherId: ObjectId,
+    options?: OptionsBase
+  ): Promise<AffectedRecords>;
+}
+
+export interface Dispatcher {
+  id?: ObjectId;
+  /** The type of event the Dispatcher will respond e.g 'mail' or 'task' */
```
AaronTrippaers, line 75: Can we look for example  of actual events that fire: like `user_deleted` for example. The `mail` or `task` are action types, that makes a tad bit confusing.

AaronTrippaers (approved):
You can also fix the comment in the next PR, it is not breaking anyway 😄

PR #801, opened 2023-09-18
SVCS-2025 Add findAll and findAllIterator

AaronTrippaers:


src/services/dispatchers/dispatchers/types.ts
```diff
@@ -19,6 +20,30 @@ export interface DispatchersService {
    */
   find(options?: OptionsWithRql): Promise<PagedResult<Dispatcher>>;
 
+  /**
```
AaronTrippaers, line 23: Double check the jsdocs 😄  Did not find good examples for these 😬

src/services/dispatchers/dispatchers/index.ts
```diff
@@ -40,6 +41,14 @@ export default (
       return await query(options);
     },
 
+    async findAll(this: DispatchersService, options) {
+      return findAllGeneric<Dispatcher>(this.find, options);
```
chad-pelly, line 45: I think both instances of `this.find` should be `query` 🤔

src/services/dispatchers/dispatchers/types.ts
chad-pelly: Regarding the comments on the interfaces for the Dispatcher findAll / findAll iterator:
https://github.com/ExtraHorizon/javascript-sdk/blob/358ec06cb9aad18774d7dfac1f607c6d399fa948/src/services/auth/oidc/providers/types.ts

chad-pelly (commented):
Just two very small comments

src/services/dispatchers/dispatchers/index.ts
AaronTrippaers: Copied from: https://github.com/ExtraHorizon/javascript-sdk/blob/358ec06cb9aad18774d7dfac1f607c6d399fa948/src/services/tasks/tasks.ts but there we don't have query, so I will change it 😄

src/services/dispatchers/dispatchers/types.ts
```diff
@@ -19,6 +20,31 @@ export interface DispatchersService {
...
+  findAll(options?: OptionsWithRql): Promise<Dispatcher[]>;
+
+  /**
+   * ## Retrieve a paged list of Dispatchers
+   *
+   * **Global Permissions:**
+   * - `VIEW_DISPATCHERS` - Allows a user to view Dispatchers
+   *
+   * @param options {@link OptionsWithRql} - Add filters to the requested list
+   * @returns An iterator for the queried providers {@link FindAllIterator FindAllIterator<Dispatcher>}
```
chad-pelly, line 43: `providers` -> `Dispatchers` 😜

src/services/dispatchers/dispatchers/types.ts
```diff
@@ -19,6 +20,31 @@ export interface DispatchersService {
...
+   * @throws {@link NoPermissionError} when the user doesn't have the required permissions to execute the function.
```
chad-pelly, line 44: This is also not related to functions and we decided to not add the errors here I think 🤔

chad-pelly: Same comments in the findAll 😂

chad-pelly (commented):
Few minor things still 😂

chad-pelly (approved):

PR #808, opened 2023-10-13
SVCS-2336 Added skip count operator to the RQL builder

chad-pelly:


src/rql/index.ts
```diff
@@ -129,6 +129,22 @@ export interface RQLBuilder {
    */
   excludes: (field: string, ...expressions: RQLString[]) => RQLBuilder;
 
+  /**
+   * @description skipCount() Skips the document counting step of a request to increase performance.
```
wwaaijer-exh, line 133: I belief we refer to "records" rather the "documents" in other RQL operators 🤔

wwaaijer-exh (commented):
Nice, just a little comment on the phrasing 👌

src/rql/index.ts
chad-pelly: i like that ALOT 🆗

wwaaijer-exh (approved):

PR #811, opened 2023-10-20
Feature/svcs 2340 double encode all values in the sdk for utf 8 support

chad-pelly:


src/rql/index.ts
```diff
@@ -110,16 +129,17 @@ export function rqlBuilder(rql?: RQLString | string): RQLBuilder {
   };
 
   function processQuery(operation: string, value: string) {
-    if (value.includes(' ')) {
+    if (!shouldDoubleEncode && value.includes(' ')) {
```
AaronTrippaers, line 132: Will you move this warning to the doubleEncode function in the next PR? 
I feel like there we can check if the encoding of the value differs from the actual value when `shouldDoubleEncode` is false and then give the error message, which is way more broad and better feedback.

PR #813, opened 2023-10-23
SVCS-2340 Added logic to double encode RQL values by default

chad-pelly:


src/rql/index.ts
```diff
@@ -109,17 +135,30 @@ export function rqlBuilder(rql?: RQLString | string): RQLBuilder {
...
+  function encodeValue(value: string) {
+    return encodeURIComponent(value).replace(
+      /[-_.!~*'()]/g,
+      c => `%${c.charCodeAt(0).toString(16).toUpperCase()}`
+    );
+  }
+
+  function processQuery(operation: string, value: string) {
     returnString = returnString
       .concat(returnString.length > 0 ? '&' : '')
```
AaronTrippaers, line 158: Annoying question: but it is problematic that we use the shorthand here? Not sure if the & breaks something 🤔

chad-pelly: Great notice! 🙏

src/rql/index.ts
```diff
@@ -20,8 +26,26 @@ export function rqlParser(rql: string): RQLString {
...
-export function rqlBuilder(rql?: RQLString | string): RQLBuilder {
-  let returnString = rql && rql.charAt(0) === '?' ? rql.substr(1) : rql || '';
+export const rqlBuilder: RqlBuilderFactory = (
+  input: RQLBuilderInput
+): RQLBuilder => {
+  let returnString = '';
+  let shouldDoubleEncode = rqlBuilder.doubleEncodeValues;
+
+  let rqlString: RQLBuilderString;
+  if (typeof input === 'object') {
```
wwaaijer-exh, line 36: (minor) nitpicky, `null` is also considered an object. If `null` is supplied, `const { doubleEncode, rql } = input;` will fail, I belief 🤔

wwaaijer-exh (commented):
Looking good, just some small comments 👌

src/rql/index.ts
chad-pelly: Tested with Aäron and this does not negatively effect the outcome thus it will be left as it is! 👍

src/rql/index.ts
```diff
@@ -109,17 +120,59 @@ export function rqlBuilder(rql?: RQLString | string): RQLBuilder {
     },
   };
 
-  function processQuery(operation: string, value: string) {
-    if (value.includes(' ')) {
-      console.log(
-        'A space has been detected while building the rql, please be aware that problems can arise'
-      );
+  // Resolve the input for a rql builder
+  function determineInput() {
```
wwaaijer-exh, line 124: (opinion, no need to change!) If we accept `input` as a parameter, this function could be removed from the inner scope of the "rqlBuilderFactory". In my opinion making it easier to process what is going on within this function and more clear in its purpose when it is used at line 13.

chad-pelly: Absolutely I had initially done it this way, but I thought it perhaps was nicer to have all the variables it was using being scoped within it's own function. If that's your preference, Im actually happy to extract it back out.

wwaaijer-exh: I'm a fan 👌 Nice "pure" function, easy to follow and reason about

wwaaijer-exh (approved):

PR #814, opened 2023-10-24
SVCS-2370 Updated changelog

chad-pelly:


CHANGELOG.md
```diff
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [v8.0.0]
+
+### Changed
```
chad-pelly, line 10: Should we mention the users and data service?

AaronTrippaers: I don't think so, the user service is not changed regarding utf-8 support. And the data service should ony differ from the other node service with the `<` and the `>`.

CHANGELOG.md
```diff
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
...
+
+### Changed 
+- Double encoding values provided to the rql builder is enabled by default
+- Example query: `rqlBuilder().eq(name, '< value >').build()`
+  - Current behaviour: `?eq(name,%253C%2520value%2520%253E)`
+  - Old behaviour: `?eq(name,< value >)`
+
+### Added
+- An option to disable double encoding for all RQL operations with `rqlBuilder.doubleEncodeValues = false`
+- An option to disable double encoding for a single RQL operation `rqlBuilder({ doubleEncodeValues: false }).eq(name, '< value >').build()`
```
AaronTrippaers, line 18: Is the skipCount not also added?

chad-pelly: Yup, I also added the ticket to the SDK release 🤦

CHANGELOG.md
```diff
@@ -5,6 +5,20 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [v8.0.0]
+
+### Changed 
+- Double encoding values provided to the rql builder is enabled by default
+- Example query: `rqlBuilder().eq(name, '< value >').build()`
+  - Current behaviour: `?eq(name,%253C%2520value%2520%253E)`
+  - Old behaviour: `?eq(name,< value >)`
```
wwaaijer-exh, line 14: I belief it may be stressed more that this is a breaking change. I think we could urge developers to check their project for the use of encoding mechanisms like `encodeURIComponent()` before upgrading and/or setting the `rqlBuilder.doubleEncodeValues = false`

wwaaijer-exh (approved):

PR #816, opened 2023-10-25
Added PoC of OAuth1 signing to match the authentication service

chad-pelly:


src/http/oauth1.ts
```diff
@@ -99,24 +102,60 @@ export function createOAuth1HttpClient(
...
+      oauth_token: tokenData.token,
+      oauth_version: '1.0',
+    };
+
+    const signatureParameters = {
+      ...searchParameters,
+      ...omit(parameters, ['realm', 'oauth_realm', 'oauth_signature']),
+    };
+
+    const signature = oauthSign.sign(
```
chad-pelly, line 131: All of the code is written to match the input of these boys
https://github.com/ExtraHorizon/auth-service/blob/dev/src/lib/oauth1Signature.ts

AaronTrippaers:
Actual implementation is now implemented

PR #817, opened 2023-10-25
Feature/svcs 2425 make should double encode default to false and downgrade to version 7.7.0

chad-pelly:


tests/unit/rql.test.ts
```diff
@@ -308,17 +316,20 @@ describe('rql string builder', () => {
...
     const firstRql = rqlBuilder().in('firstNames', [firstName]).build();
-    expect(firstRql).toStrictEqual(`?in(firstNames,${firstName})`);
+    expect(firstRql).toStrictEqual(`?in(firstNames,${doubleEncodedFirstName})`);
 
     const secondRql = rqlBuilder().out('lastNames', [lastName]).build();
-    expect(secondRql).toStrictEqual(`?out(lastNames,${lastName})`);
+    expect(secondRql).toStrictEqual(`?out(lastNames,${doubleEncodedLastName})`);
 
     // Reset the default global value
     rqlBuilder.doubleEncodeValues = true;
```
AaronTrippaers, line 335: Does this make sense at the end of the test? I believe we can remove it. There is another one exactly the same.

chad-pelly: Correct, i missed them.

tests/unit/rql.test.ts
```diff
@@ -308,17 +316,20 @@ describe('rql string builder', () => {
   });
 
   it('Should globally override the double encoding of RQL values', () => {
```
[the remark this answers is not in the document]
chad-pelly: This is a general override, the other tests override the global then the RQL builder is overridden with options.

AaronTrippaers (approved):

PR #826, opened 2023-11-08
Add v7.7.1 to the changelog

AaronTrippaers:


CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [v7.7.1]
+
+### Fixed
+- Double Encoding of RQL values is now also supported in `OAuth1` clients
```
wwaaijer-exh, line 11: (minor) Might it be interesting to mention we reimplemented the oAuth1 signature generation for this? If we made a mistake and receive oAuth signature errors, maybe users would make the connection that upgrading to this version is to blame

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [v7.7.1]
+
+### Fixed
```
chad-pelly, line 10: Isn't this now Added? 🫣

AaronTrippaers: I still see this as fixed since you could already set the double encoding to true even if it was an oAuth1 client. It was just broken.

chad-pelly: For me we added a new feature being a new method to sign OAuth1 requests AND we fixed OAuth1 signature exceptions when using double encoded characters

Just my two cents, I'll approve regardless.

wwaaijer-exh: Now that we've agreed to go for a 7.8.0 version this could indeed be listed as `Changed`?

wwaaijer-exh (approved):

PR #828, opened 2023-11-09
Stop dependabot from making PR's for now

wwaaijer-exh:
We want to have more control over this and a better suite of tests for more confidence in dependency upgrades


chad-pelly (approved):
🙏

PR #832, opened 2023-11-21
SVCS-2224 Add option to skip custom properties normalization

AaronTrippaers:


src/http/interceptors.ts
```diff
@@ -45,9 +50,18 @@ export const camelizeResponseData = ({
     ['arraybuffer', 'stream'].includes(config.responseType ?? '') ||
     config?.interceptors?.skipCamelizeResponseData
       ? data
-      : camelizeKeys(data),
+      : camelizeKeys(data, config.skipKeyNormalizationForProperties || []),
```
AaronTrippaers, line 53: Are we sure we want to keep this name? I still find it hard to reason with a negative (`skip`) in the name? 
Can we call it `normalizeCustomProperties` and default to `false` ?

src/http/oauth1.ts
```diff
@@ -204,6 +204,8 @@ export function createOAuth1HttpClient(
    */
   return Object.defineProperty(
     {
+      skipCaseNormalizationForCustomProperties:
+        options.skipCaseNormalizationForCustomProperties !== false,
```
AaronTrippaers, line 208: Check is like this because of the negative in the name. `undefined` should be true.

src/http/utils.ts
```diff
@@ -37,15 +37,29 @@ function is(Ctor, value) {
...
   if (is(Object, obj)) {
     return Object.keys(obj).reduce((memo, key) => {
-      if (is(Object, obj[key]) && !is(Date, obj[key])) {
-        return { ...memo, [fn(key)]: recursiveRenameKeys(fn, obj[key]) };
+      if (
+        is(Object, obj[key]) &&
+        !is(Date, obj[key]) &&
+        !skipKeys.includes(key)
+      ) {
+        const keys = skipKeys
```
AaronTrippaers, line 58: Add a comment here to what is exactly happening with the keys.

src/services/helpers.ts
```diff
@@ -117,3 +124,44 @@ export function addPagersFn<T>(
     next,
   };
 }
+
+export function addCustomPropertiesToConfig(
```
AaronTrippaers, line 128: I know this function can be written shorter but I went for super expressive here 😄

src/services/helpers.ts
```diff
@@ -117,3 +124,44 @@ export function addPagersFn<T>(
...
+  customProperties: string[],
+  httpInstance: HttpInstance,
+  requestOptions: OptionsBase
+) {
+  const camilizedCustomProperties = customProperties.map(camelize);
+  const snakifiedCustomProperties = customProperties.map(decamelize);
```
AaronTrippaers, line 134: Add a comment here why we do the camelizing and decamelizing here.

src/services/types.ts
```diff
@@ -61,6 +61,8 @@ export type OptionsBase = {
   headers?: Headers;
   /** Only passed to the GET requests. Will retry 4 times on 500 errors */
   shouldRetry?: boolean;
+  /** Set to false when needed */
```
AaronTrippaers, line 64: Improve comment

src/services/profiles/profiles.ts
```diff
@@ -2,14 +2,28 @@ import type { HttpInstance } from '../../types';
...
 
 export default (
   client: HttpClient,
   httpAuth: HttpInstance
 ): ProfilesService => ({
   async find(options) {
-    return (await client.get(httpAuth, `/${options?.rql || ''}`, options)).data;
+    const customProperties = [
+      'data.custom_fields',
+      'data.groups.custom_fields',
```
AaronTrippaers, line 18: The `.*` is now implicit at the end of every string . `custom_fields` is still converted to `customFields`

src/http/oauth1.ts
chad-pelly: Isn't it kind of weird that we have `skipKeyNormalizationForProperties` and `skipCaseNormalizationForCustomProperties`? 

Aren't they the same thing, but targeting different things? or do the names key vs case mean something different?

chad-pelly (commented):
I'm really struggling to ingest this all at once with the complexity and certainly how non stupid this code is. Just a few questions to get started.

src/http/interceptors.ts
```diff
@@ -45,9 +50,18 @@ export const camelizeResponseData = ({
...
 });
 
+export const decamilizeRequestData = (
+  data,
+  httpRequestConfig?: HttpRequestConfig
+) =>
+  decamelizeKeys(
```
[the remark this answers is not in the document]
AaronTrippaers: Agree but this PR is complicated enough, so I don't really want to change it now.

src/http/interceptors.ts
chad-pelly: You mean default it to the empty array? 🤔

src/http/oauth1.ts
AaronTrippaers: `skipCaseNormalizationForCustomProperties` is a setting we expose where we try to explain to the customer about which type of properties this is about. The custom ones. That is just a boolean they can turn off and on again.

`skipKeyNormalizationForProperties` is internal to the sdk which says to skip the normalization for the properties in the array that are given to it. They don't necessarily have to be custom, it is whatever we provide it with.

src/http/interceptors.ts
AaronTrippaers: Comment is in the wrong place 😒  Going to put it in the correct place.

tests/unit/profiles/profilesService.test.ts
```diff
@@ -49,6 +49,126 @@ describe('Profiles Service', () => {
...
+    };
+
+    const rql = rqlBuilder().build();
+    nock(`${host}${PROFILES_BASE}`)
+      .get('/')
+      .reply(200, createPagedResponse(profile));
+
+    const res = await sdk.profiles.find({
+      rql,
+      skipCaseNormalizationForCustomProperties: false,
```
AaronTrippaers, line 87: Are we sure want to keep calling this with the `skip` making it a double negative.
We could call it `normalizationCasingForCustomProperties` and default it to false instead of true, that would be easier to reason with.

src/services/helpers.ts
```diff
@@ -117,3 +124,49 @@ export function addPagersFn<T>(
...
+        ...snakifiedCustomProperties,
+      ],
+    };
+  }
+
+  // If nothing is set on the request, we need to check the global settings
+  // If globally turned off, don't skip on any properties
+  if (httpInstance.skipCaseNormalizationForCustomProperties === false) {
+    return requestOptions;
+  }
```
wwaaijer-exh, line 162: (opinion/discussion) Could this be simplified if the toggle and the custom fields were different variables?

A `customProperties` variable  would always be added to the request object if applicable to the endpoint. (This would also aid the future RQL thingy we've discussed)

A `skipCaseNormalizationForCustomProperties` is also in the request object to toggle the feeding of the customProperties to decamalize/camelize or not. Which could be done at the last moment in `decamilizeRequestData` or `camelizeResponseData`

Keeping the definition there are "custom properties" in an endpoint and how that influences the behavior with extra toggles separate.

I might be missing stuff though, happy to discuss

AaronTrippaers: 💯  love this idea

src/services/profiles/groups.ts
```diff
@@ -7,18 +8,26 @@ export default (
...
   async create(profileId, requestBody, options) {
+    const customProperties = ['custom_fields'];
+
     return (
-      await client.post(httpAuth, `/${profileId}/groups`, requestBody, options)
+      await client.post(
+        httpAuth,
+        `/${profileId}/groups`,
+        requestBody,
+        addCustomPropertiesToConfig(customProperties, httpAuth, options)
```
wwaaijer-exh, line 18: (no need to change, just a thought) I think this is good, but I do feel like it could be as easy as `{ ...options, customFields }` and "hide" the complexity of the transformation the custom fields to the relevant "transform..." functions.

But that might be a step too far. Was trying to simplify this in my head, as this piece of code is going to end up in a lot of places

src/services/profiles/profiles.ts
```diff
@@ -32,11 +46,28 @@ export default (
   },
 
   async create(requestBody, options) {
-    return (await client.post(httpAuth, '/', requestBody, options)).data;
+    const customProperties = ['custom_fields', 'groups.custom_fields'];
```
wwaaijer-exh, line 49: Would it make sense to add just a single test for each of these methods to prove they're working? (No need to add all the test you added for `find` to all these other methods though)

src/services/profiles/groups.ts
```diff
@@ -30,12 +39,13 @@ export default (
...
 
   async removeFields(profileId, groupId, requestBody, options) {
+    const customProperties = ['custom_fields'];
     return (
       await client.post(
         httpAuth,
         `/${profileId}/groups/${groupId}/remove_fields`,
         requestBody,
-        options
+        addCustomPropertiesToConfig(customProperties, httpAuth, options)
```
AaronTrippaers, line 48: Rename `addCustomPropertiesToConfig` to `addCustomPropertiesToRequestOptions`

src/http/interceptors.ts
```diff
@@ -45,9 +50,23 @@ export const camelizeResponseData = ({
...
     config?.interceptors?.skipCamelizeResponseData
       ? data
-      : camelizeKeys(data),
+      : camelizeKeys(
+          data,
+          config?.normalizeCustomPropertyCasing ? [] : config?.customProperties
+        ),
 });
 
+export const decamilizeRequestData = (
```
chad-pelly, line 59: I think this should be `decamelizeRequestData` and also should it be `snakifyRequestData` or `snake'nmakeRequestData`.... 🐍

chad-pelly (commented):
Some more thoughts!

src/http/interceptors.ts
```diff
@@ -45,9 +50,23 @@ export const camelizeResponseData = ({
     ['arraybuffer', 'stream'].includes(config.responseType ?? '') ||
     config?.interceptors?.skipCamelizeResponseData
       ? data
-      : camelizeKeys(data),
+      : camelizeKeys(
+          data,
+          config?.normalizeCustomPropertyCasing ? [] : config?.customProperties
```
AaronTrippaers, line 55: Mark `normalizeCustomPropertyCasing` as deprecated

src/http/interceptors.ts
AaronTrippaers: Create a ticket for this version to rename `decamelize` -> `snakify`

src/http/utils.ts
```diff
@@ -37,15 +37,35 @@ function is(Ctor, value) {
...
-        return { ...memo, [fn(key)]: recursiveRenameKeys(fn, obj[key]) };
+      // Check that `obj[key]` is an object
+      // If it is date or `skipKeys` includes the current key, don't continue the recursion
+      if (
+        is(Object, obj[key]) &&
+        !is(Date, obj[key]) &&
+        !skipKeys.includes(key)
+      ) {
+        // Filter all keys in `skipKeys` that start with the current `key` before the first `.`
+        // Remove the current `key` from those keys before going into the next iteration.
```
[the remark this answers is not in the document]
AaronTrippaers: - Explain what the `skipKeys` does conceptually: e.g: these keys are excluded from the transformation. 
- Rename keys to partial keys or something else

src/http/utils.ts
```diff
@@ -73,15 +93,17 @@ export function decamelize(string: string): string {
 }
 
 export function camelizeKeys(
-  object: Record<string, unknown>
+  object: Record<string, unknown>,
+  keysToSkip?: string[]
```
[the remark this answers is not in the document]
AaronTrippaers: Align `keys` and `properties` everywhere
Align `skip` and `ignore` everywhere

Prefer `keys` over `properties` and `ignore` over `skip`

src/http/interceptors.ts
```diff
@@ -1,8 +1,13 @@
 import { AxiosInstance, AxiosRequestConfig } from 'axios';
```
AaronTrippaers, line 1: - Check with with FibriCheck if this actually solves their problem
- Check with Fibricheck if the doubleEncoding property solve their problem

Focus should be that it is a breaking version but you can fallback to the previous behaviour.

src/http/interceptors.ts
AaronTrippaers: https://extrahorizon.atlassian.net/browse/SVCS-2573

src/services/profiles/profiles.ts
```diff
@@ -32,11 +43,28 @@ export default (
   },
 
   async create(requestBody, options) {
-    return (await client.post(httpAuth, '/', requestBody, options)).data;
+    const customKeys = ['custom_fields', 'groups.custom_fields'];
```
AaronTrippaers, line 46: @wwaaijer-exh i have added these based on the swagger file but I think this is not needed, since they can not be added to the profile in a creation, they can probably also not be returned?

src/services/profiles/profiles.ts
AaronTrippaers: Gave it go but with the current state of the tests it would be a large and even more annoying change, will add this as a question to the questions document to see if we want to do this for next PR's and maybe make a new one for the profiles.

src/services/profiles/profiles.ts
wwaaijer-exh: You're right, it should never be returned. But this seems like the best future proof and consistent thing to do 👌

src/http/interceptors.ts
```diff
@@ -86,7 +105,11 @@ export const transformResponseData = ({
     ['arraybuffer', 'stream'].includes(config?.responseType ?? '') ||
     config?.interceptors?.skipTransformResponseData
       ? data
-      : recursiveMap(mapDateValues, config?.url?.startsWith(DATA_BASE))(data),
```
AaronTrippaers, line 89: Still doubting the removal of of the `config?.url?.startsWith(DATA_BASE)` here, it might make a difference on raw request to the data service 🤔

wwaaijer-exh (commented):
Ohhh yeah, sooooo close I can feel it 👌

src/http/interceptors.ts
AaronTrippaers: @wwaaijer-exh for the camelize we indeed decided to do that. For this one we decided something else completely ignoring the raw calls: see the last sentence on this questions answer: https://extrahorizon.atlassian.net/wiki/spaces/EHPE/pages/235765917/SDK+-+8.0.0+-+Questions#%3Acheck_mark%3A--Are-we-considering-the-data-transform-as-well%3F

AaronTrippaers: This should solve the problem: Do add it to the changes here?

AaronTrippaers: https://github.com/ExtraHorizon/javascript-sdk/commit/b2b7eef06199713d4c6d039a1a1c14ebb3f10ef2

src/http/utils.ts
```diff
@@ -5,29 +5,30 @@ export function mapObjIndexed(fn, object): Record<string, unknown> {
...
-    return Array.isArray(obj)
-      ? obj.map(recursiveMap(fn, skipData))
-      : mapObjIndexed((value, key) => {
-          if (typeof value !== 'object') {
-            return fn(value, key);
-          }
+    // Filter all keys in `ignoreKeys` that start with the current `key` before the first `.`
+    // Remove the current `key` from those keys before going into the next iteration.
+    const partialIgnoreKeys = ignoreKeys
+      .filter(k => k.split('.')[0] === key)
```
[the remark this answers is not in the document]
chad-pelly: Same for line 72? Again I could be wrong?

src/services/helpers.ts
```diff
@@ -117,3 +119,32 @@ export function addPagersFn<T>(
     next,
   };
 }
+
+export function setCustomKeysConfigurationInRequestConfig(
+  httpInstance: HttpInstance,
+  requestOptions?: HttpRequestConfig
+) {
+  const customRequestKeys =
+    requestOptions?.customRequestKeys || requestOptions?.customKeys;
```
chad-pelly, line 128: (question) Is it not a possibility that we want to extend the customKeys with customRequestKeys? The same with the response, rather than redeclaring keys? Maybe it makes it too complex to deduce what your set of keys is?

chad-pelly (dismissed):
Just a few questions nothing blocking 👌

src/http/utils.ts
```diff
@@ -36,16 +37,45 @@ export const recursiveMap =
...
+ * @param ignoreKeys An array of keys to ignore during the renaming process.
+ * @returns An object with keys renamed according to the provided function.
+ */
+export function recursiveRenameKeys(
+  fn: { (arg: string): string },
+  obj,
+  ignoreKeys: string[] = []
+) {
+  // If ignoreKeys includes the '*', we don't need to convert anything in this object
+  if (ignoreKeys.includes('*')) return obj;
```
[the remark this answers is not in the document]
AaronTrippaers: Both functions work the other way around. I can align them with a small re-factor in this PR as well if you want 😄 

In `recursiveMap` it immediatly check if it is object and other wise skips it.
In `recursiveKeyNameMap` it first only checks the wildcard and then it checks the objects after the array. It was already like that and I did not really change that .

src/http/interceptors.ts
AaronTrippaers: raw calls, for `sdk.data...` it is indeed correct if we add the custom keys

src/http/utils.ts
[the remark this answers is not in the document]
AaronTrippaers: Correct , I will change it. The startsWith is indeed a bit more efficient.

src/services/helpers.ts
AaronTrippaers: That is a design choice, both are just as easy to implement and both have advantages. 

If I see this, I would expect the `customKeys` are overwritten by the `customRequestKeys`. That is why I went with the overwrite. 
```
{
    customKeys: ['b`],
    customRequestKeys: ['a']
}
```

src/http/utils.ts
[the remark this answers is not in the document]
wwaaijer-exh: `startsWith` would result in false positives though, no?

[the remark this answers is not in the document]
AaronTrippaers: @wwaaijer-exh How would it result in false positives? I think both are the same 🤔

src/http/utils.ts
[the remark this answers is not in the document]
chad-pelly: Nah if you didn't change it YOLO, just again a question to check 😄

src/services/helpers.ts
chad-pelly: Just a thought and maybe any future questions can be solved with a little comment over `setCustomKeysConfigurationInRequestConfig` ?

wwaaijer-exh (approved):

PR #834, opened 2023-11-24
SVCS-2460 Added double encoding to the 8.0.0 changelog

chad-pelly:


CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [8.0.0]
+
+### Changed
```
AaronTrippaers, line 10: Should there a mention somewhere that these are breaking?

AaronTrippaers: We did this in earlier versions: https://github.com/ExtraHorizon/javascript-sdk/blob/ebbc3600e370d791457c4e98eba7a84ddee334e7/CHANGELOG.md#breaking-changes-1

chad-pelly: Hmmmm CHANGED is breaking:

These are the guidelines we state we follow: https://keepachangelog.com/en/1.1.0/

AaronTrippaers: Then we are using changed wrong in all the minor and patched version 😬 
I would rather be consistent with ourselves at this point then with the standard.

wwaaijer-exh: I'm leaning towards following the standard, but the standard also states it should be "painfully clear" that something is breaking 😄 Would it make sense to add a bold **BREAKING** in front of the item? In my opinion it should be more clear this is breaking, but no hard stance on how 🤔

wwaaijer-exh (commented):
Also nice! 👌

CHANGELOG.md
chad-pelly: We can meet in the middle for sure 🤗

AaronTrippaers (approved):

PR #833, opened 2023-11-24
Feature/svcs 2428 default rql builder.double encode values to true and mark double encode values as deprecated

chad-pelly:

PR #835, opened 2023-11-24
SVCS-2581 Added double encoding upgrade to the README

chad-pelly:


README.md
```diff
@@ -103,6 +103,15 @@ find(
 ): Promise<PagedResult<Schema>>;
 ```
 
+## 8.0.0 Upgrade Guide
+
+### Double Encoding
+- Ensure that instances of encoding values for the rql builder such as `encodeURIComponent()` are removed
+- Disable double encoding for a single rql builder operation `rqlBuilder({ doubleEncodeValues: false })` where double encoding is not desired
+
+#### NOT ADVISED
```
AaronTrippaers, line 112: Same comment: https://github.com/ExtraHorizon/javascript-sdk/pull/833#discussion_r1404160131

README.md
```diff
@@ -103,6 +103,15 @@ find(
 ): Promise<PagedResult<Schema>>;
 ```
 
+## 8.0.0 Upgrade Guide
+
+### Double Encoding
```
wwaaijer-exh, line 108: I know I said I didn't want a book, but I do belief we can elaborate a bit more like you did in the inline comment in the deprecation. I would even suggest putting most of that message here and just a small description and reference to here.

Would it make sense to start of by stating something like "From now on values supplied to RQL operations are now automatically encoded to be safe to be used in RQL. Therefore any (URL) encoding of these values currently implemented should be removed."?

And suggesting if this is not feasible you could disable globally and optionally enable on each call. But like you said, that should be something temporary.

Maybe this would make sense to be discussed like documentation as well 😅

wwaaijer-exh (approved):

README.md
```diff
@@ -103,6 +103,30 @@ find(
...
+## 8.0.0 Upgrade Guide
+
+### Double encoding RQL values 
+From version 8.0.0 RQL values are now [double encoded](https://docs.extrahorizon.com/extrahorizon/additional-resources/resource-query-language-rql#double-encoding-of-special-characters) by default when using the rql builder, to support the use of special characters in rql operations.
+
+In your existing application it is likely that you have encoded values provided to the rql builder using methods such as `encodeURIComponent()`. To support the automatic double encoding of special characters, where applicable, these instances of encoding should be removed.
+- Before 8.0.0: `rqlBuilder().eq('diagnosis', encodeURIComponent('Hypertension - STAGE 1'))`
+- After 8.0.0: `rqlBuilder().eq('diagnosis', 'Hypertension - STAGE 1')`
+
+Any use of an rql builder that does not need to support special characters may remain as it is.
```
AaronTrippaers, line 115: What is the purpose of this sentence? 🤔 
I don't think it says a lot.

README.md
```diff
@@ -103,6 +103,30 @@ find(
...
+
+#### Reverting double encoding for a single rql builder
+You may disable double encoding for an rql builder by setting the option `rqlBuilder({ doubleEncodeValues: false })`
+- Before 8.0.0: `rqlBuilder().eq('diagnosis', encodeURIComponent('Hypertension - STAGE 1'))`
+- After 8.0.0: `rqlBuilder({ doubleEncodeValues: false }).eq('diagnosis', encodeURIComponent('Hypertension - STAGE 1'))`
+
+#### Reverting double encoding for all rql builders
+The automatic double encoding of values can be reverted to behavior before 8.0.0 using the snippet `rqlBuilder.doubleEncodeValues = false`. This should normally exist alongside the declaration of your Extra Horizon SDK instance at the root of your application.
```
AaronTrippaers, line 123: `automatic double encoding` and `double encoding` are used a bit at random

AaronTrippaers (approved):
Looks good, two minor comments but up to you if you still want to fix them.

README.md
chad-pelly: We also have to inform users that they only need to make changes in the scenarios that they are using special characters no?

e.g. `rqlBuilder().eq('diagnosis', 'Hypertension')` would require no changes.

README.md
chad-pelly: Double encoding is used when referencing the rql builder as it's object
Automatic double encoding is used when not referencing the object to ensure the knowledge that the user isn't in control of it.

PR #836, opened 2023-11-27
SVCS-2590 Set customKeys for data-service

AaronTrippaers:


src/services/data/documents.ts
```diff
@@ -10,14 +10,16 @@ export default (
   httpAuth: HttpInstance
```
AaronTrippaers, line 10: Since it is a node-service that only has the response conversion, we don't need to adapt end points that only have custom request data

AaronTrippaers: Or do we want to set them anyway for if we move in the future?

src/services/data/documents.ts
```diff
@@ -10,14 +10,16 @@ export default (
...
 ): DataDocumentsService => {
   function partialApplyFind(schemaId) {
```
chad-pelly, line 12: Comment on the existing code not on you:

Following the way that we 'should implement' the get endpoints for a service this is WAY off.  I wonder if we should make a ticket to fix this up in the future. Looking at the code for the `addPagersFn`, `findAllGeneric`, etc.. methods, it's actually hard to tell if this is working as intended or by 'accident'.

```ts
  async function query(schemaId, options) {
    const customResponseKeys = ['data.data'];
    const baseConfig = options?.headers ? {headers: options.headers} : {};

    const { data } = await client.get(httpAuth, `/${schemaId}/documents${options?.rql || ''}`, { ...baseConfig, customResponseKeys });
    return data;
  }
```

AaronTrippaers: This can be an improvement ticket for the SDK. Do you want to make it?

src/services/data/documents.ts
wwaaijer-exh: 👌 If we don't do this now I think a note in the ticket to move to the requestInterceptor would be helpful though

src/services/data/schemas.ts
```diff
@@ -24,7 +24,10 @@ export default (
 ): DataSchemasService => {
   async function find(options) {
     const result: PagedResult<Schema> = (
-      await client.get(httpAuth, `/${options?.rql || ''}`, options)
+      await client.get(httpAuth, `/${options?.rql || ''}`, {
+        ...options,
+        customResponseKeys: ['data.properties', 'data.statuses'],
```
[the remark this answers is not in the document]
wwaaijer-exh: The argument Friday was to limit the changes needed everywhere.

Little improvements like the one mentioned above by Chad or for instance moving the adding of rql a layer down (WHY IS `${options?.rql || ''` EVERYWHERE?! sorry), could bring the complexity and meat of these endpoints and their option objects down quite a lot. These custom fields have more right to be here than rql and stuff like `addTransitionHelpersToSchema`, imho.

Lots of text but I do not have a strong opinion here though! Either way is fine.

Long way of saying: I agree, but if it is "extra work" I think we would be solving the wrong part of the problem?

src/services/data/documents.ts
chad-pelly: I'll make a ticket 🤗

src/services/data/schemas.ts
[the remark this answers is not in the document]
chad-pelly: Agreed ^^ it's just once we commit to this we have 5+ more tickets to do, I just wanted to make sure we're aligned on how we want to do it rather than regretting it in a few days.

I agree with the rql options and MANY other things.

src/services/data/documents.ts
chad-pelly: https://extrahorizon.atlassian.net/browse/SVCS-2618

src/services/data/documents.ts
AaronTrippaers: https://extrahorizon.atlassian.net/browse/SVCS-2583

PR #837, opened 2023-11-28
Feature/svcs 2589 add custom keys for all the endpoints in the configurations service

chad-pelly:


src/services/configurations/general.ts
```diff
@@ -7,7 +7,18 @@ export default (
   httpAuth: HttpInstance
 ): ConfigurationsGeneralService => ({
   async get(options) {
-    return (await client.get(httpAuth, '/general', options)).data;
+    const { data } = await client.get(httpAuth, '/general', {
+      ...options,
+      customResponseKeys: [
+        'data.*',
```
AaronTrippaers, line 13: The `/*` is unnecessary if there is a name before it  It probably will work but you don't really have to put it

src/services/configurations/users.ts
```diff
@@ -7,7 +7,16 @@ export default (
   httpAuth: HttpInstance
 ): ConfigurationsUsersService => ({
   async get(userId, options) {
-    return (await client.get(httpAuth, `/users/${userId}`, options)).data;
+    const { data } = await client.get(httpAuth, `/users/${userId}`, {
+      ...options,
+      customResponseKeys: [
+        'data.*',
+        'staffConfiguration.*',
+        'patientConfiguration.*',
```
AaronTrippaers, line 15: I think this is wrong, if I look at the swagger this should be `staffConfiguration.data` and `patientConfiguration.data`

AaronTrippaers (commented):
Looks good! Found 1 small mistake.

tests/__helpers__/configuration.ts
```diff
@@ -35,6 +62,25 @@ export const groupConfig = {
...
+};
+
+export const customGroupConfigResponse = {
+  ...customGroupConfig,
+  id: randomHexString(24),
+  updateTimestamp: '2018-04-24T11:57:44.525Z',
+  creationTimestamp: '2018-04-24T11:57:44.525Z',
+};
+
 export const groupConfigResponse = {
```
AaronTrippaers, line 84: Should this one be up together with the groupConfig still?

tests/unit/configurations/generalService.test.ts
```diff
@@ -51,6 +52,20 @@ describe('Configuration: General Service', () => {
     expect(res.patientConfiguration).toBeDefined();
   });
 
+  it('Should not transform custom data in the response', async () => {
```
AaronTrippaers, line 55: I would add in response **of the get request**, since there are also other end points tested here.

src/services/configurations/general.ts
chad-pelly: I find a little conflicting the object is customResponseKeys.
I would expect that it wants specific keys hence the wildcard to re-enforce it's all keys in this object. 🤔 

It works both ways, but it's up to whatever people want.

tests/__helpers__/configuration.ts
chad-pelly: I'm not sure I see or understand what you mean? 😅

AaronTrippaers: You have put two object between `groupConfig` (line 53)  and `groupConfigResponse` (line 84) while those objects itself belong together.

src/services/configurations/general.ts
AaronTrippaers: It does not really matter, but it would have been good if this came up in the first PR. Now we are already inconsistent with ourselfves because the profile service and the data service don't do it like that.

AaronTrippaers: I agree that this makes more sense though

tests/__helpers__/configuration.ts
chad-pelly: Updated the file to be more clear

src/services/configurations/general.ts
chad-pelly: What is the point of the `*` though as a wildcard if we don't consume it? 🤔 

We would almost never user it in the root of something because of timestamps right? 🤔

wwaaijer-exh: I also think it looks nicer 👍 Although I also agree the extra work isn't ideal, we all want to move to this "`.*` syntax"?

AaronTrippaers: @chad-pelly the localization service needs the wildcard, that is why I added it. 

@wwaaijer-exh if we all want to move to the .* syntax someone should go back through the profile and data service and change it. 

Also then we have the question wether `custom_fields` should still work as it does know  or wether that should just not change `custom_fields` in itself.

wwaaijer-exh: Can we make this a "questions document question" and/or have a quick call soon(-ish)? I feel like we're not all fully aligned. Having this discussion async and spread out over a day is most likely not going to very efficient 🤔

wwaaijer-exh (dismissed):

AaronTrippaers (approved):

wwaaijer-exh (approved):

PR #838, opened 2023-11-28
SVCS-2597 Add mail service custom keys

AaronTrippaers:


src/services/mails/mails.ts
```diff
@@ -11,7 +11,12 @@ export default (client: HttpClient, httpAuth: HttpInstance): MailsService => ({
   },
 
   async find(options) {
```
AaronTrippaers, line 13: e2e tests are coming

AaronTrippaers: https://github.com/ExtraHorizon/exh-end-2-end-test/pull/45/files

wwaaijer-exh (approved):

PR #841, opened 2023-11-28
SVCS-2622 Added comment to custom keys options

chad-pelly:


chad-pelly:
I see one super quirky edge case that we couldn't solve with the current implementation, probably doesn't exist but maybe me writing it down triggers something in Wilco's big ol' brain.

```
{
	id: "1234567897654321"
	"exhField": {
		"required_underscored_field": "ya wat m8?"
	},
	user_defined_field: {
	}
}
```

src/http/types.ts
```diff
@@ -25,8 +25,36 @@ export type HttpRequestConfig = AxiosRequestConfig & {
     retryCondition: (error: HttpResponseError) => boolean;
   };
   normalizeCustomData?: boolean;
+  /**
+   * **Providing a key with an object value**:
+   * - The provided key is not ignored for case conversion on a request and response
+   * - The child keys of the object are ignored for case and date conversion on a request and response
+   *
+   * **Providing a key with a non-object value**:
+   * - The provided key is ignored for case and date conversion on a request and response
```
AaronTrippaers, line 34: I don't think this is correct. If you provided a key for non object value, not happens.

src/http/types.ts
```diff
@@ -25,8 +25,36 @@ export type HttpRequestConfig = AxiosRequestConfig & {
     retryCondition: (error: HttpResponseError) => boolean;
   };
   normalizeCustomData?: boolean;
```
AaronTrippaers, line 27: Should there be a mention in the comments that this holds only true is `normalizeCustomData` is false?

src/http/types.ts
chad-pelly: This feels pretty counter-intuitive... 😅

chad-pelly: I also find it very strange that 

const {...options, customKeys: ['users_key_and_string_value'] } would do absolutely nothing. 😅

AaronTrippaers: The function really works as if there is a `.*` attached to any input. 
It ignores all fields below the key that you are specifying.  If the key you specify has a value of a string, there are no further fields. 

It is really important to see these properties as a means to an end to help ourselves with the current state of the SDK. it is not exposed to any of our users and it should not cover all possible options that exist, just enough to help us out here 😄

chad-pelly: But this means if we have a user defined field that is a just a key string in the root, we can't ignore it without ignoring everything in the root?

AaronTrippaers: The case you are stating is not solvable with the custom keys anyway. Since if it is a key with a user defined name, we can't list it in the custom keys because we don't know what users will choose 😄

chad-pelly: Doesn't have to be user defined, it could be our own value that we don't camel case

AaronTrippaers: The main pupose of doing this was:
- All our own fields should be camelCased.
- All user defined fields should stay as they are.

So that should not be a real problem. 

We will see quickly see if there are still actually edge cases that are not working since we are now using the custom keys for all services.

wwaaijer-exh:
> I see two super quirky edge cases

That would indeed not be solvable. Luckily I really dislike (with a passion) mixing "user" and "system" fields at the same level. The only place I know of we did this is in the data service, There we've reserved the `id` field on objects within arrays. But (luckily) that is not affecting this case 😄

AaronTrippaers (approved):

PR #844, opened 2023-12-01
Add custom keys for the task service

AaronTrippaers:


src/services/tasks/index.ts
```diff
@@ -1,15 +1,40 @@
...
+import api from './api';
+import apiRequests from './apiRequests';
+import functions from './functions';
+import logs from './logs';
+import schedules from './schedules';
 import tasks from './tasks';
 import { TASKS_BASE } from '../../constants';
 import { TasksService } from './types';
 
-export const tasksService = (httpWithAuth: HttpInstance): TasksService => {
```
AaronTrippaers, line 7: Moved all of it here from the tasks functions. Probably should not have done it in this PR 😅

wwaaijer-exh: It was a bit confusing, but nice change though 🎉

wwaaijer-exh (approved):

PR #847, opened 2023-12-06
SVCS-2565 Added interceptors  to unauthenticated client and tests

chad-pelly:


src/http/client.ts
```diff
@@ -47,10 +50,12 @@ export function createHttpClient({
     );
   }
 
-  http.interceptors.response.use(
-    camelizeResponseData,
-    typeReceivedErrorsInterceptor
-  );
+  http.interceptors.response.use(null, retryInterceptor(http));
```
chad-pelly, line 53: Could be we don't want these interceptors?

```js
  http.interceptors.response.use(null, retryInterceptor(http));
  http.interceptors.response.use(null, typeReceivedErrorsInterceptor);
  ```

Comment (AaronTrippaers) [reply src/http/client.ts:53]: The `typeReceivedErrorsInterceptor` was already there, so I would keep that one.

Comment (AaronTrippaers) [reply src/http/client.ts:53]: I don't think the `retryInterceptor` can hurt here?  If the user sets the amount of retries in the config, he will expect it to happen, also if the call is not authenticated.

File: tests/unit/auth/unauthenticated.ts [hunk 2]
```diff
@@ -0,0 +1,83 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
AaronTrippaers, line 1: Not sure if there is a better way but this file feels a bit out of place 🤔 
It is in the auth service tests but it tests the configurations service. 
Would `tests/unit/client/client.test.ts` be a better place to put these tests?

tests/unit/auth/unauthenticated.ts
```diff
@@ -0,0 +1,83 @@
...
+import nock from 'nock';
```
AaronTrippaers, line 2: Would it also make sense to adapt the `it('should register a new user', async () => {` test because that is the endpoint that will always use this client?

AaronTrippaers (approved):
Nothing blocking, just some thoughts on the tests

wwaaijer-exh (dismissed):
Nice 👌 Same comments as Aäron, nothing to add

src/http/client.ts
```diff
@@ -2,6 +2,9 @@ import axios from 'axios';
 import { HttpClientConfig, HttpInstance } from '../types';
```
wwaaijer-exh, line 2: Having had some time to think about it, this does change a few extra things I belief 🤔 The non authenticated `http` client is used in some other calls that would now also have their timestamp changed. **Might be nothing**, but just want to check if we're aware of this. User Service seems to be fine, just looking at the oAuth1/2/proxy internal calls now, which do seem to be affected

chad-pelly: I don't see any changes besides the one interface for the OAuth1TokenData boi? Did you notice anything I missed? 🤔

wwaaijer-exh: I indeed think that is all 🤔

tests/unit/client/client.test.ts
```diff
@@ -30,4 +52,70 @@ describe('HttpClient', () => {
...
+    });
+
+    expect(response).toStrictEqual({
+      ...authenticationResponse,
+      creationTimestamp: new Date(authenticationResponse.creationTimestamp),
+      updateTimestamp: new Date(authenticationResponse.updateTimestamp),
+    });
+  });
+
+  it('Should transform dates in the response when creating an account', async () => {
```
wwaaijer-exh, line 73: (NO NEED TO CHANGE! 😄) Just to clarify, from my side the intent was to adjust the existing test for the `sdk.users.createAccount` in `tests\unit\users\usersService.test.ts` called `should register a new user`.

I know this is not from the perspective of the changes but rather from the outside. I'm guessing those different views might have caused some confusion. (Same goes for the `authencticate` and `confirmMfa` method tests of the oAuth1 client btw)

wwaaijer-exh (approved):
Noice, looking good 👌

AaronTrippaers (approved):

PR #848, opened 2023-12-07
SVCS-2561 Use different package for hmac sign

AaronTrippaers:

PR #849, opened 2023-12-07
Fix dependabot errors

AaronTrippaers:


package.json
```diff
@@ -50,7 +50,7 @@
     "@rollup/plugin-json": "^4.1.0",
     "@rollup/plugin-node-resolve": "^13.1.3",
```
AaronTrippaers, line 51: All errors are fixed except for: https://github.com/ExtraHorizon/javascript-sdk/pull/848/files and the one where we need to a major upgrade on axios.

PR #867, opened 2024-02-23
SVCS-2936 Added methods to work with the user verification settings

wwaaijer-exh:


src/services/users/index.ts
```diff
@@ -5,6 +5,8 @@ import httpClient from '../http-client';
 import globalRoles from './globalRoles';
 import groupRoles from './groupRoles';
 import health from './health';
+import { userSettingsService } from './settings';
```
chad-pelly, line 8: (Stupid) Could also just be `Settings` or `SettingsService` because we're already in the users context here?

chad-pelly (approved):
Only 1 silly thought! 💪

chad-pelly (approved):
NOICE! 🥳

PR #869, opened 2024-02-26
SVCS-2933 Added the changes for the forgot password pin code mode

wwaaijer-exh:


src/services/users/types.ts
```diff
@@ -916,33 +946,79 @@ export interface UsersService {
...
+   * The token should be used to complete the password reset via `exh.users.validatePasswordReset`.
+   *
+   * By default the email send to the user is the email template configured by `passwordResetEmailTemplateId`.
+   * The template receives a 40 hexadecimal character hash in the `content.reset_hash` variable.
+   *
+   * If enabled, a pin code can be used rather than a hash.
+   * The pin code mode must be enabled by the `enablePinCodeForgotPasswordRequests` verification setting.
+   * To use the pin code mode, the `mode` field can be set to `pin_code`.
+   * Then the email send to the user is the email template configured by `passwordResetPinEmailTemplateId`.
+   * The pin code template receives a 8 digit pin code in the `content.pin_code` variable.
```
wwaaijer-exh, line 984: I know these texts are not ideal. The goal here was just to provide the facts. A nice to read text should be coming up later 😬 (Spelling mistakes and logical corrections are more then welcome though)

chad-pelly (approved):
👌

src/services/users/types.ts
chad-pelly: It's more than good enough 😂

src/services/users/users.ts
```diff
@@ -157,17 +157,25 @@ export default (
...
+        clientOptions.params.email = data;
+      } else {
+        clientOptions.params.email = data.email;
+
+        if (data.mode) {
+          clientOptions.params.mode = data.mode;
+        }
+      }
+
+      const response = await userClient.get(http, '/forgot_password', clientOptions);
```
chad-pelly, line 176: This doesn't need to change at all, but it feels like the request parameters are buried a little bit here for me, although they seem like the most important part 🤔 

Maybe you hate the below syntax more... definitely doesn't need to change, just the readers eyes 👍 
```JS
const response = await userClient.get(http, '/forgot_password',{...options, params});
```

wwaaijer-exh: 👍 My reasoning was that if we ever add the ability to add params in the options object, we'll override them. But that is not really reasonable to think about now and like your argument better. Changed

chad-pelly (approved):
👍

PR #871, opened 2024-02-27
SVCS-2712 Automatically refresh oAuth2 tokens before expiring

wwaaijer-exh:


src/http/oauth2.ts
```diff
@@ -113,11 +124,24 @@ export function createOAuth2HttpClient(
...
   httpWithAuth.interceptors.response.use(transformKeysResponseData);
 
-  async function setTokenData(data: TokenDataOauth2) {
+  /**
+   * - Adds a creationTimestamp to the token
+   * - Notifies the user
+   * - Stores the token to authenticate future requests
+   * - Returns the token
+   */
+  async function dateAndSetTokenData(data: TokenDataOauth2) {
```
wwaaijer-exh, line 133: Name can be discussed for sure, started off with `setAndHydrateTokenData`.
(But maybe the whole concept of adding the timestamp and returning is not great. See other comment)

src/http/oauth2.ts
```diff
@@ -140,8 +164,8 @@ export function createOAuth2HttpClient(
         } :
         {}
     );
-    await setTokenData(tokenResult.data);
-    return tokenResult.data;
+
+    return await dateAndSetTokenData(tokenResult.data);
```
wwaaijer-exh, line 168: To prevent adding the `creationTimestamp` and extra return statement everywhere in the 3 spots, I've consolidated it in the function which was already used. Happy to discuss if this feels too complex

CHANGELOG.md
```diff
@@ -24,6 +24,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `activationPinEmailTemplateId`
   - `reactivationPinEmailTemplateId`
   - `passwordResetPinEmailTemplateId`
+- Improved oAuth2 access token refreshing
+  - If we're able to estimate the expiry time of a token, the token is refreshed before it expires
```
wwaaijer-exh, line 28: (probably best to read after reviewing the rest of the PR 😅)

Currently this means if you're restoring tokens, from something like local storage, the initial request is still going to fail first and then refresh and retry. This would reduce the scope of the re-upload issue, but not completely solve it.

Alternatively we could accept both the `expiresIn` and `creationTimestamp` next to the `accessToken` and `refreshToken` when constructing the SDK. (As we are returning these fields) This felt too much to add in this PR or could be a little overkill.

The example mentioned in the docs simply initializes the SDK without tokens and explicitly generates new tokens by authenticating with the refresh token grant: https://github.com/ExtraHorizon/javascript-sdk/blob/dev/gitbook/examples/authentication.md#snippet-for-authentication-flow-oauth2
(This would never have re-uploading issue)

This would have the issue, as we cannot estimate the time the access token will expire:
```ts
createOAuth2Client({
  host: "...",
  clientId: "...",
  accessToken: "...",
  refreshToken: "...",
});
```

This would be the extended proposal:
```ts
createOAuth2Client({
  host: "...",
  clientId: "...",
  accessToken: "...",
  refreshToken: "...",
  expiresAt: "...",
  creationTimestamp: "...",
});
```
(This is in the same form as we return the values, meaning a simple dump/restore from local storage would be enough)

Happy to discuss, create a ticket to pick this up later, or open another PR for this!

CHANGELOG.md
```diff
@@ -40,6 +44,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - `exh.auth.applications.deleteVersion` now calls the correct endpoint
+- Automatic oAuth2 access token refreshing now properly uses client credentials
```
wwaaijer-exh, line 47: This was from the previously fixed ticket

CHANGELOG.md
chad-pelly: Love the fix, I think it's more than adequate 🤗 

I have two further questions:
- Should it be us or our clients responsibility to 'silently login' to address your first comment?
- If a user is uploading a file that takes 5 minutes, and the token expires within the first 3 minutes, the request is not interrupted right?

Perhaps we can discuss the first thing this afternoon? 🤔

wwaaijer-exh: Further questions:
1) In the current implementation, the users. Only relevant if they'll want to prevent a retry, the normal refresh after error mechanism still keeps working.
2) Indeed, not interrupted. Tokens are checked at the start of the upload.
Quick discussion in the office sounds good 👍

chad-pelly: After a 'quick discussion' the:
```
  expiresAt: "...",
  creationTimestamp: "...",
  ```
  
  Are definitely something that can be added now 👍

chad-pelly (dismissed):
👌

AaronTrippaers (approved):

PR #887, opened 2024-03-26
SVCS-2668 SVCS-2175 Fix a problem with oAuth1 and uploads

wwaaijer-exh:


src/http/oauth1.ts
```diff
@@ -94,7 +94,6 @@ export function createOAuth1HttpClient(
     ...config,
     headers: {
       ...config.headers,
-      'Content-Type': 'application/json',
```
wwaaijer-exh, line 97: Did find a bug preventing uploads with react native (both iOS and Android) it was only with oAuth1 because of this header being set (which it doesn't need to). Not a problem for web/node.js it seems 😅

AaronTrippaers (approved):

PR #889, opened 2024-03-29
SVCS-737 SVCS-906 Example consistency, no more sdk and fibricheck.com

wwaaijer-exh:


chad-pelly (commented):
Shouldn't this work be done on the Gitbook branch? 🤔 

I feel like this is maybe will cause the conflict thingies we discussed a few weeks back when we merge them after the PC?

wwaaijer-exh:
@chad-pelly Wanted to keep the changes together. Was a bit biased by you guys not being in and did not want to create too many different PR's. But you're absolutely right 😄

Removed the change from this PR, but made a bit of a mess by just now immediately merging the extracted changes the Gitbook docs. You can few the changes here though: https://github.com/ExtraHorizon/javascript-sdk/commit/d7cabe2c99d2076a3a8c753706fa008c61d7ab75
 
As an upside, discovered we should spent some more time on explaining the oAuth 2 authoriztion flow. The flow was mentioned to be specific for FibriCheck, but is also available to other parties. (as seen in the Gitbook docs commit) https://extrahorizon.atlassian.net/browse/SVCS-3065

chad-pelly (approved):
Awesome work! 💪

PR #890, opened 2024-04-02
Add strong typing to RQLBuilder

tran-simon:
When executing a query like

```typescript
  await exh.data.documents.find(rqlBuilder().in('id', ids).build())
```

the `fields` parameter was previously typed as a string.

This PR adds an optional generic parameter to enable strong typing:

```typescript
  // TS will show an error if `id` is not a field of `MyObject`
  await exh.data.documents.find<MyObject>(rqlBuilder().in('id', ids).build())
```

### DIFF
--- a/src/rql/index.ts
+++ b/src/rql/index.ts
@@ -25,19 +25,19 @@ export function rqlParser(rql: string): RQLString {
  * @see https://developers.extrahorizon.io/guide/rql.html
  * @returns
  */
-export const rqlBuilder: RqlBuilderFactory = (
+export const rqlBuilder: RqlBuilderFactory = <T extends Record<string, any>>(
   input: RQLBuilderInput
-): RQLBuilder => {
+): RQLBuilder<T> => {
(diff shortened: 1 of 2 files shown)


tran-simon:
I realised that this will not work properly with deep fields:

```
  await exh.data.documents.find<Document<MyObject>>(rqlBuilder().in('data.myField', values).build())
```

We will need a type utility to strongly type a dot notation path, something like

https://medium.com/xgeeks/typescript-utility-keyof-nested-object-fa3e457ef2b2

```
export type KeyPath<T extends object | undefined> = T extends object
  ? {
    [Key in keyof T & string]: T[Key] extends object | undefined
      ? Key | `${Key}.${KeyPath<T[Key]>}`
      : Key;
  }[keyof T & string]
  : never;
 ```
 
 I'll work on this later

wwaaijer-exh:
Hey Simon, thank you for this!

We have been working on a similar prototype, but cases like you described increase the scope of this quite rapidly. I'm very interested to see what you end up with.

Issue #698, opened 2022-11-08
the data.documents.unlinkGroups does not have the possibility to provide no groupid's


ThomasReyskens:
## Description
I want to unlink all groups, but this requires an empty object. Currently, this is not possible because this is not optional. -> `unlinkGroups(schemaId: string, documentId: string, requestBody: {  groupIds: string[] };`

simply changing this to 
`unlinkGroups(schemaId: string, documentId: string, requestBody: {  groupIds: string[] } | {};` should suffice I guess!

wwaaijer-exh:
Fixed in `8.2.0`

PR #897, opened 2024-04-04
SVCS-446 exh.data.documents.unlinkGroups request body is now optional

wwaaijer-exh:


src/services/data/types.ts
```diff
@@ -769,8 +769,8 @@ export interface DataDocumentsService {
   unlinkGroups(
     schemaIdOrName: ObjectId | string,
     documentId: ObjectId,
-    requestBody: {
```
AaronTrippaers, line 772: In my opinion it is safer and better ux to keep both required and create an extra `unlinkAllGroups` function.

But that is more because I don't really like the design of the end point as it is than due to this fix.

AaronTrippaers (dismissed):

src/services/data/types.ts
wwaaijer-exh: I agree! But since it is describing the endpoint I feel this should exist (at least) in this form. Would it make sense to add a ticket to create the extra helper at some point?

AaronTrippaers (approved):

src/services/data/types.ts
wwaaijer-exh: Ahh okay, I didn't fully understand you position. I agree keeping the need for an empty object helps to prevent you shooting yourself in the foot. Thomas also mentioned this in his original issue (#698). Fixed

PR #906, opened 2024-04-18
Correct invalid types

tran-simon:
- Interfaces `TransitionActionLinkUserFromData` and `TransitionActionLinkGroupFromData` had the wrong `type` field
- Set return type of `findById` and `findFirst` to potentially undefined


wwaaijer-exh (approved):
Looks good 🎉 We'll be publishing a version later next week

Issue #687, opened 2022-10-12
RqlBuilder or() not working correctly when providing multiple statements


ThomasReyskens:
## Description
RqlBuilder's `or` doesn't work when adding multiple clauses. The server replies w/ an invalid rql response.
The builder should transform the incoming string to an `and` clause to fix this

## Steps to Reproduce
1. build and rql string with multiple clauses in an or statement
2. submit request
3. error

## Code example

```typescript
const rql = rqlBuilder()
    .or(
      rqlBuilder().eq('status', 'reviewd').intermediate(),
      rqlBuilder().eq('status', 'analyzed').eq('data.indicator','normal').intermediate()
    ).build();
```

## Specifications

- Version: 7.1.1-dev.212

SalinaExh:
https://extrahorizon.atlassian.net/browse/SVCS-449

wwaaijer-exh:
How I believe the RQL builder is building this:
`or(status=reviewed,status=analyzed&data.indicator=normal)`
How I believe the RQL build should build this:
`or(status=reviewed,eq(status=analyzed,data.indicator=normal))`

wwaaijer-exh:
Current workaround:
```js
rqlBuilder()
  .or(
    rqlBuilder().eq('status', 'reviewd').intermediate(),
    rqlBuilder().and(
      rqlBuilder().eq('status', 'analyzed').intermediate(),
      rqlBuilder().eq('data.indicator','normal').intermediate()
    ).intermediate()
  ).build();
```

wwaaijer-exh:
Similar issue reported by Pieter some time ago:
```js
// Not working
  rqlBuilder().contains('patient_enlistments', rqlBuilder()
    .in('group_id', visibleGroupsIds)
    .eq('expired', 'false')
    .intermediate())
  .build();
```
Proposed workaround:
```js
rqlBuilder().contains(
  'patient_enlistments',
  rqlBuilder().in('group_id', visibleGroupsIds).intermediate(),
  rqlBuilder().eq('expired', 'false').intermediate()
).build();
```

wwaaijer-exh:
Fixed in `8.3.0`

PR #911, opened 2024-05-08
SVCS-449 `rql.intermediate` now wraps multiple expressions in an `and`

wwaaijer-exh:
Solving #687

Multiple commits to fix the linting errors (test files are not linted, but errors do show up in editor, slowly fixing them)


chad-pelly (dismissed):
🪨 ⭐

Issue #935, opened 2024-09-09
`documents.unlinkUsers` request body type should have `userIds` optional


tran-simon:
## Description

The `unlinkUsers` function is typed as:

```typescript
  unlinkUsers(
    schemaIdOrName: ObjectId | string,
    documentId: ObjectId,
    requestBody: {
      userIds: Array<ObjectId>;
    },
    options?: OptionsBase
  ): Promise<AffectedRecords>;
```

https://github.com/ExtraHorizon/javascript-sdk/blob/dev/src/services/data/types.ts#L820-L827

However, in the jsdoc it says that

> **Not** specifying the `userIds` array will **unlink all** users from the document.

You currently cannot not specify the `userIds` array, as it is typed as required

wwaaijer-exh:
Hey Simon, thanks for the report and the PR! We'll pick it up in the beginning of next month with some other changes we have planned

wwaaijer-exh:
Was fixed in 8.5.0 🎉

PR #939, opened 2024-10-03
Don't require the request body to be set for unlinkGroups/unlinkUsers

wwaaijer-exh:
Extending on the point made in #935, the UX for this is not great.

After discussion:

Deprecated:
```js
exh.data.documents.unlinkGroups(schemaId, documentId, {})
exh.data.documents.unlinkGroups(schemaId, documentId, { groupIds: ['58d1047f52faff0007680834'] })
```

Introduced:
```js
exh.data.documents.unlinkAllGroups(schemaId, documentId)
exh.data.documents.unlinkGroups(schemaId, documentId, ['58d1047f52faff0007680834']) // Overloads current method
```

### DIFF
--- a/src/services/data/documents.ts
+++ b/src/services/data/documents.ts
@@ -151,11 +151,13 @@ export default (
     },
 
     async unlinkGroups(schemaIdOrName, documentId, requestBody, options) {
+      const data = requestBody || {};
+
       return (
         await client.post(
           httpAuth,
           `/${schemaIdOrName}/documents/${documentId}/unlinkGroups`,
-          requestBody,
+          data,
           options
         )
       ).data;
@@ -173,11 +175,13 @@ export default (
     },
 
     async unlinkUsers(schemaIdOrName, documentId, requestBody, options) {
+      const data = requestBody || {};
+
       return (
         await client.post(
           httpAuth,
           `/${schemaIdOrName}/documents/${documentId}/unlinkUsers`,
-          requestBody,
+          data,
           options
         )
       ).data;
--- a/src/services/data/documents.ts
+++ b/src/services/data/documents.ts
@@ -150,17 +150,21 @@ export default (
       ).data;
     },
 
-    async unlinkGroups(schemaIdOrName, documentId, requestBody, options) {
-      const data = requestBody || {};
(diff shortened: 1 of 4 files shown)


AaronTrippaers (dismissed):

src/services/data/documents.ts
```diff
@@ -151,11 +151,13 @@ export default (
     },
 
     async unlinkGroups(schemaIdOrName, documentId, requestBody, options) {
+      const data = requestBody || {};
+
       return (
         await client.post(
           httpAuth,
           `/${schemaIdOrName}/documents/${documentId}/unlinkGroups`,
-          requestBody,
```
AaronTrippaers, line 158: I approved but I have the same hesitation as https://github.com/ExtraHorizon/javascript-sdk/pull/897#discussion_r1551541330

chad-pelly: I also kind of agree with Aäron it feels weirdly implicit that if an `requestBody` argument is not provided that it unlinks everything.

I do belieb that this operation should be a separate specific function and the existing behaviour for this stay the same as it would seem quite easy to make this mistake?

wwaaijer-exh: Oh damn, I completely forgot we already had this discussion. I agree that there is a risk, but also interesting we're at the same discussion again. I will gladly introduce separate methods as I belief the `{}` is just bad UX / design. However, I'm currently a little stuck regarding naming this without introducing breaking changes. I'll have a think

wwaaijer-exh: What about leaving the possibility like today:
```js
exh.data.documents.unlinkGroups(schemaId, documentId, {})
exh.data.documents.unlinkGroups(schemaId, documentId, { groupIds: ['58d1047f52faff0007680834'] })
```
But deprecating that ☝️
And overloading the `unlinkGroups` function with straight array usage and introducing `unlinkAllGroups`:
```js
exh.data.documents.unlinkAllGroups(schemaId, documentId)
exh.data.documents.unlinkGroups(schemaId, documentId, ['58d1047f52faff0007680834'])
```
This way it is harder to shoot yourself in the foot, but the old behavior is still possible and no weird new name needs to be found for the "specific group unlinking"-case. Traded for some extra complexity on our side dealing with the overloaded function call.

chad-pelly (approved):
Solid! �

AaronTrippaers (approved):

PR #941, opened 2024-10-22
Add TRANSITION_DOCUMENTS permission

AaronTrippaers:


src/services/users/types.ts
```diff
@@ -386,8 +387,6 @@ export enum GlobalPermissionName {
   /** @deprecated Not in use any longer */
   MANAGE_SERVICE_ALERTS = 'MANAGE_SERVICE_ALERTS',
   /** @deprecated Not in use any longer */
-  TRANSITION_DOCUMENTS = 'TRANSITION_DOCUMENTS',
-  /** @deprecated Not in use any longer */
```
AaronTrippaers, line 390: did we ever have this permission? Putting it back feels icky if it is deprecated.

wwaaijer-exh: If I remember correctly, initially when creating the SDK this list was extracted from the dev environment admin role. Meaning it included a few permissions that never made it to production. This was confusing and I assume we didn't want to make breaking changes at the time and just marked them as deprecated 🤔 I wouldn't think too much of it 😄

wwaaijer-exh (approved):
🚀

PR #942, opened 2024-11-14
Mention undefined as a return value for some template finding methods

wwaaijer-exh:


src/services/templates/types.ts
```diff
@@ -106,23 +106,20 @@ export interface TemplatesService {
   /**
    * Find By Id
    * @param id the Id to search for
-   * @param rql an optional rql string
```
wwaaijer-exh, line 109: Also removed this since it is incorrect (and little to no added value imho)

chad-pelly (approved):

PR #946, opened 2024-11-20
EXH-192 Support triggering a transition by name

wwaaijer-exh:


chad-pelly (approved):
(Minor) Would it make sense that we add a transition by name helper here?

We already have `findTransitionIdByName` as well as all the other `byName` helpers so it isn't a far stretch for me to consider

PR #952, opened 2025-01-14
Feature/exh 223 sdk update the data service interfaces to match the updated swagger permissions

chad-pelly:


src/services/data/types.ts
```diff
@@ -564,106 +583,189 @@ export interface DataDocumentsService {
...
   ): Promise<Document<OutputData, CustomStatus>>;
+
   /**
-   * Request a list of documents
-   *
-   * ReadMode on schema is set to 'default' (or the property is not set at all on the schema):
-   *
-   * Permission | Scope | Effect
-   * - | - | -
-   * none | | See your own documents
```
chad-pelly, line 574: Either this was wrong or the current Swagger documentation is wrong 🤔 

`"default"` | Translates to `["linkedUsers","linkedGroupStaff"]` relational access mode

src/services/data/types.ts
```diff
@@ -673,27 +775,98 @@ export interface DataDocumentsService {
...
+   * **Partially** update the selected document, provide `null` as a value to clear a field.
+   * <br>
+   * <br>
    *
-   * Permission | Scope | Effect
+   * ## Access via permissions
+   * Regardless of how the access modes (described below) are set, a user is always able to perform an operation on a document if they are assigned a specific permission.  This permission can come from a global role of the user or a staff enlistment role the user has in the group of the document.
+   * Permission | Scopes | Effect
    * - | - | -
-   * none | | Update your own documents
```
chad-pelly, line 694: Same as above

src/services/data/types.ts
wwaaijer-exh: I think (and maybe hope 😅) that this is correct, but (very) awkwardly phrased. The "linkedUsers" part is covered by this row of the table here. The "linkedGroupStaff" part is covered by "none | `staff enlistment` | See the documents belonging to the group" mentioned below this row.

src/services/data/types.ts
```diff
@@ -755,26 +981,55 @@ export interface DataDocumentsService {
...
    *
    * Start a transition manually for the selected document where the conditions of a manual transition are met.
    *
-   * Either the `id` or the `name` of a transition must be provided.
-   * If both are provided, they must match the same transition, otherwise a `ResourceUnknownError` will be thrown.
-   *
-   * If `name` is provided, multiple transitions could match.
-   * In this case, the first matching transition will be triggered:
-   * - The first transition with a `fromStatuses` matching the current status of the document
-   * - And all the transition its `conditions` are met
```
wwaaijer-exh, line 768: Not sure if it matters much, but this description/context is now removed right?

wwaaijer-exh (approved):

PR #958, opened 2025-03-18
Feature/exh 302 sdk implement notifications service v2

chad-pelly:


.github/workflows/buildAndPublish.yml
```diff
@@ -7,6 +7,7 @@ on:
     branches:
       - master
       - dev
+      - feature/*
```
wwaaijer-exh, line 10: Maybe we can remove these before continuing to dev now?

PR #974, opened 2025-07-08
Added the new endpoints for notifications service v2

wwaaijer-exh:


src/services/notificationsV2/users/types.ts
```diff
@@ -52,20 +130,22 @@ export interface NotificationV2UserService {
...
-  * # Interface
+  * ## Interface
   * @returns NotificationV2User[]
   */
   findAll(options?: OptionsWithRql): Promise<NotificationV2User[]>;
 
   /**
   * # Request notification settings for a specific user
   *
+  * This will only return a user notification settings object if the user has notification settings set up.
```
AaronTrippaers, line 141: I think this method might be strange from a user perspective now, can we remove it or can it be in use already?

AaronTrippaers (dismissed):

src/services/notificationsV2/users/types.ts
chad-pelly: I actually kind of like it, but I would phrase it differently:
"Unlink the `getById` endpoint, if a user settings object does not exist, it will not be created"

Or something like that

AaronTrippaers (approved):

PR #999, opened 2026-01-08
Feature/exh 1131 sdk update dependencies project structure

chad-pelly:


package.json
```diff
@@ -37,7 +37,7 @@
     "fflate": "0.8.2",
     "form-data": "4.0.4",
     "platform-specific": "1.1.0",
-    "qs": "6.12.0",
+    "qs": "6.14.1",
```
AaronTrippaers, line 40: (minor/silly) We have this comment lying around: 
{ decodeDotInKeys: false } // Without this oAuth1 signatures will break, opened an issue in qs: https://github.com/ljharb/qs/issues/500 . This is fixed since by @wwaaijer-exh , Maybe we remove the comment now?

AaronTrippaers (dismissed):

package.json
chad-pelly: That's fair, but do we want to remove the comment or the whole line? As the setting is actually returned to the default? 🤔

wwaaijer-exh: If it's not immediately clear what we want to do, lets just create a ticket for it an pick it up in the next upcoming release (for the template service)

wwaaijer-exh (approved):

package.json
AaronTrippaers: A ticket is also good 👌  I don't think there is a wrong choice here.
I would choose between:
- removing the comment and the code
- keep the code and adapt the comment to say that is there to make sure it never changes again

AaronTrippaers (approved):

package.json
chad-pelly: Perfect 👌 

There's a ticket and we can pick it up with the templates service, I'll focus on just patching things for now:
https://extrahorizon.atlassian.net/browse/EXH-1152

PR #1004, opened 2026-02-05
Add templates service V2 support

AaronTrippaers:


src/services/templatesV2/types.ts
```diff
@@ -0,0 +1,190 @@
...
+} from '../types';
+
+export interface TemplateV2Out {
+  id?: ObjectId;
+  name?: string;
+  description?: string;
+  properties?: Record<string, TypeConfiguration>;
+  outputs?: Record<string, string>;
+  creationTimestamp?: Date;
+  updateTimestamp?: Date;
```
wwaaijer-exh: I think we can mark `id`, `name`, `outputs` and the timestamps as non optional? I know we don't do this correctly in a lot of places, but might be nice to start doing it

src/services/templatesV2/health.ts
```diff
@@ -0,0 +1,13 @@
+import type { HttpInstance } from '../../types';
+import { ResultResponse, Results } from '../../types';
+
+export default (client, http: HttpInstance) => ({
```
wwaaijer-exh, line 4: In the spirit of making things better going forward (again I know we've done this a lot), do we really need this to be a separate file? (Do we even need the health endpoint is also a question, but i guess there are some (edge) cases)

src/services/templatesV2/templatesV2.ts
```diff
@@ -0,0 +1,76 @@
+import { rqlBuilder } from '../../rql';
```
chad-pelly, line 1: I'm not sure where to put this comment, because accounts for the file:

I think we might want to consider referring to the 'latest' version of writing these services e.g. the notificationV2 as I feel syntactically they are quite different and it makes sense to keep making them in the same vein.

Happy to discuss this in a call

src/services/templatesV2/health.ts
wwaaijer-exh: Looking at some other stuff we did for the notification service V2 I think there is a lot to pick up, maybe we can discuss in the after noon and do a few refactor PR's

wwaaijer-exh (dismissed):

src/services/templatesV2/types.ts
```diff
@@ -0,0 +1,190 @@
...
+  OptionsWithRql,
+  OptionsBase,
+} from '../types';
+
+export interface TemplateV2Out {
+  id?: ObjectId;
+  name?: string;
+  description?: string;
+  properties?: Record<string, TypeConfiguration>;
+  outputs?: Record<string, string>;
```
chad-pelly: I think it would be nice for properties and ouputs, like the notification service V2 we could make them generics? 

It could also happily be another ticket 🫡

src/services/templatesV2/types.ts
```diff
@@ -0,0 +1,190 @@
...
+  creationTimestamp?: Date;
+  updateTimestamp?: Date;
+}
+
+export interface TemplateV2In {
```
chad-pelly: In the same vain as the above comment would it be more consistent
`TemplateV2Creation` for `In` and `TemplateV2<T>` for out?

src/services/templatesV2/types.ts
AaronTrippaers: How do you see the generic(s) being used and helpful in this case? I am struggling to see it here.

src/services/templatesV2/templatesV2.ts
AaronTrippaers: I have adapted it, hope it is more in line now 😄

wwaaijer-exh (dismissed):

chad-pelly (approved):

Issue #957, opened 2025-02-18
Task Service: Most function endpoints are missing


wwaaijer-exh:
Get/post/update/delete/... are all actions on the Task Service functions, but they're not implemented in the SDK yet

wwaaijer-exh:
Will be included in 8.9.0

PR #1013, opened 2026-02-06
EXH-1056 Added the  missing `exh.tasks.functions` methods

wwaaijer-exh:
Also covers [#957](https://github.com/ExtraHorizon/javascript-sdk/issues/957)


src/services/tasks/functions/index.ts
```diff
@@ -6,13 +6,55 @@ export default (
...
+
+  async update(name, body, options) {
+    const response = await client.put(httpAuth, `/functions/${name}`, body, {
+      ...options,
+      customRequestKeys: ['environmentVariables'],
+    });
+    return response.data;
+  },
+
+  async delete(name, options) {
```
AaronTrippaers: (minor) isn't `remove` used more in the sdk?

AaronTrippaers (dismissed):

tests/unit/tasks/functions.ts
```diff
@@ -0,0 +1,286 @@
...
+      nock(`${host}${TASKS_BASE}`)
+        .put(`/functions/${functionName}`, data)
+        .reply(200, { affectedRecords: 1 });
+
+      const result = await exh.tasks.functions.update(functionName, data);
+
+      expect(result).toStrictEqual({ affectedRecords: 1 });
+    });
+
+    it('Does not transform customized fields in the request', async () => {
```
AaronTrippaers: I am a bit confused here, does this test do something?

wwaaijer-exh: The idea here was to cover the "customRequestFields", but like you mentioned on Friday it doesn't do anything at the moment. If it creates confusion I rather remove the test!

src/services/tasks/functions/index.ts
wwaaijer-exh: Good call, I toke it from the task schedules, but that and the oidc providers is the only place we've used delete 😅

AaronTrippaers (approved):

PR #1028, opened 2026-02-23
Deprecate getBy in favor of findBy

AaronTrippaers:


gitbook/other/change-log.md
```diff
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [8.9.0]
 
 ### Added
-- Added the `exh.tasks.functions` methods `find`, `getByName`, `update`, `enable`, `disable` and `remove`
+- Added the `exh.tasks.functions` methods `find`, `findByName`, `update`, `enable`, `disable` and `remove`
```
AaronTrippaers, line 15: I noticed I was constantly typing findByName because that is used more in the sdk

gitbook/other/change-log.md
```diff
@@ -23,6 +23,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Deprecated
 - `exh.tasks.schedules.delete` is deprecated in favor of `exh.tasks.schedules.remove`
 - `exh.auth.oidc.providers.delete` is deprecated in favor of `exh.auth.oidc.providers.remove`
+- `exh.notificationsV2.userSettings.getById` is deprecated in favor of `exh.notificationsV2.userSettings.findById`
+- `exh.localizations.getByKeys` is deprecated in favor of `exh.localizations.findByKeys`
```
AaronTrippaers, line 27: Ensure we won't copy from a getBy function again

src/services/localizations/types.ts
```diff
@@ -153,6 +169,8 @@ export interface LocalizationsService {
    */
   remove(rql: RQLString, options?: OptionsBase): Promise<AffectedRecords>;
   /**
+   * @deprecated Use `findByKeys` instead.
+   *
```
wwaaijer-exh: To "really motivate" people to switch, we can remove the rest of the description (I think we did that for at least some of the other methods)

AaronTrippaers:
`getBy` errors when there is no entity found.
`findBy` returns `null` when there is no entity found

PR #1037, opened 2026-03-12
Finalize changelog

AaronTrippaers:


wwaaijer-exh (approved):
Ahhh yeah, good catch 🙏

PR #1051, opened 2026-05-12
EXH-1519 Update axios

AaronTrippaers:


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [8.11.0]
 
+### Fixed
+- The following vulnerabilities were resolved, but not applicable:
+  - Bumped Axios version to resolve `CVE-2026-42033`, `CVE-2026-42034`, `CVE-2026-42035`, `CVE-2026-42036`, `CVE-2026-42038`, `CVE-2026-42039`, `CVE-2026-42040`, `CVE-2026-42041`, `CVE-2026-42042` and `CVE-2026-42043`
+
```
wwaaijer-exh: I don't think this format makes a whole lot of sense for just a single package's version bump. Would it make sense to keep it as simple as some think like this?
```suggestion
- Bumped version of `axios` to resolve vulnerability warnings (vulnerable code was not in use)

```

wwaaijer-exh (approved):

PR #1067, opened 2026-06-26
Feature/exh 1721 sdk auth remove the secret values from the entities

chad-pelly:


src/services/auth/oauth2/types.ts
```diff
@@ -105,29 +105,58 @@ export interface AuthOauth2TokenService {
...
+  state?: string;
+  codeChallengeMethod?: string;
+  codeChallenge?: string;
+  authorizationCode: string;
+  expiryTimestamp: Date;
+  updateTimestamp: Date;
+  creationTimestamp: Date;
 }
 
 export interface OAuth2Authorization {
```
chad-pelly, line 128: These interface were wrong, so I updated them whilst I was here 😬

src/services/auth/oauth1/types.ts
```diff
@@ -23,7 +23,7 @@ export interface AuthOauth1Service {
    * @throws {ApplicationNotAuthenticatedError}
    * @throws {ResourceUnknownError}
    */
-  consumeSsoToken(ssoToken: string): Promise<OAuth1Token>;
+  consumeSsoToken(ssoToken: string): Promise<ConsumeSSOTokenResponse>;
```
chad-pelly, line 26: 🚨🚨🚨 WEE WOO WEE WOO 🚨🚨🚨

___ BREAKING CHANGES ALERT! ___

(multiple places) Beforehand the input type and response type were the same, however, that is no longer the case. I don't see a good way around this 😓 

Do we consider it a breaking change to change the names of the interfaces? 😬

tests/unit/auth/oauth2/index.ts
```diff
@@ -42,7 +42,6 @@ describe('Auth - OAuth2', () => {
       responseType: 'code',
       clientId: '507f191e810c19729de860ea',
       state: '',
-      scope: '',
```
chad-pelly, line 45: I don't know how scope ended up in the test and interfaces BUT I can't find a reference to it in the auth service

src/services/auth/oauth1/types.ts
chad-pelly: Maybe Wilco is right to hard remove them 😭

wwaaijer-exh: I think TypeScript is flexible enough to handle this 🤔  When the interfaces still match, these changes will not be breaking? (For example, the `OAuth1Token` interface is still exported and is a subset of `ConsumeSSOTokenResponse`, so the original interface would happily hold the new response type, i do the belief)

wwaaijer-exh: In this case I also think conceptually the `ConsumeSSOTokenResponse` is the same interface as the return value of the "authenticate" method, `TokenDataOauth1`? 🤔

src/services/auth/oauth1/types.ts
```diff
@@ -112,12 +112,28 @@ export interface SsoToken {
...
+}
+
+export interface OAuth1Token {
+  id: string;
+  applicationId: string;
+  userId: string;
+  token: string;
+  /**
+   * @deprecated `tokenSecret` will be removed in a future version.
+   * This field is no longer returned when `showSecretValues` is disabled in authentication service settings.
```
wwaaijer-exh: (discussed offline, might want to remove this line, just mentioning the removal)

src/services/auth/oauth1/types.ts
chad-pelly: As discussed in the call, in a few attempts of this I had done some inheritance gymnastics of the typings, it indeed in this revision is not breaking 🙏

wwaaijer-exh (approved):

PR #1079, opened 2026-07-15
EXH-1763 Added revised OAuth2 authorizations methods

chad-pelly:


src/services/auth/oauth2/authorizations/types.ts
```diff
@@ -0,0 +1,73 @@
...
+import { OAuth2Authorization, OAuth2AuthorizationCreation, OAuth2AuthorizationCreationResponse } from '../types';
+
+export interface OAuth2AuthorizationsService {
+    /**
+     * Create an OAuth2 authorization
+     *
+     * Permission | Scope | Effect
+     * - | - | -
+     * none | | Everyone can use this endpoint
+     */
```
chad-pelly, line 11: In the previous version we listed all the errors that can be thrown, but I find it maintenance hell 🤔 Happy to put them back though

src/services/auth/oauth2/authorizations/index.ts
```diff
@@ -0,0 +1,61 @@
+import { HttpInstance } from '../../../../http/types';
```
chad-pelly, line 1: I'm going to create a ticket to also extract the otherOAuth2 services into folders. I didn't do this for refresh tokens to keep it simple, but I really should have done.

wwaaijer-exh (approved):

